PR #49, opened 2025-12-03
feat: Add missing sensors for 12000XP and single inverter support

### LINKED ISSUE #48: 12000XP Consumption and Generator Charge SOC / Voltage Controls
Love this whole thing btw.  I am using a slight variation on you dashboards, so thatnks for those also.  Would be great to have the forecast in there (I use "solar production forcast").  

EG4 12000XP seems to be a little weird about it's consumption power.  EPS power is included in overall consumption, but not in the "now":

Image

I'd like to be able to capture the EPS power (246 W) in this instance... will probably still need a calculating sensor to add EPS to Consumption (I believe the Consumption is >0 when power is not coming from batteries, i.e. gen or grid (same things for me as I am 100% off-grid)).

From a controls perspective, the main thing I change is the Generator Start SOC/Voltage and the Generator Stop SOC/Voltage and the Generator Charge Type.  I use this to force start the generator and force stop it when I predict I will have good solar coming (no point charging to 100% if I can get by with 30% and then use the sun the next day!)

Image

btli:
## Summary

- Adds 16 missing sensor definitions that were already mapped in coordinator but lacked `SENSOR_TYPES` entries
- Updates pylxpweb dependency to >=0.3.24 which makes parallel groups optional (enables single inverter support)
- Bumps version to 3.0.0-rc.5

## New Sensors Added

**Power Sensors:**
- `power_output`: Inverter power output
- `rectifier_power`: Rectifier power (AC-DC conversion)
- `ac_couple_power`: AC coupled power
- `eps_power`: Emergency Power Supply total power
- `eps_power_l1`/`eps_power_l2`: EPS power per phase

**Voltage Sensors:**
- `bus1_voltage`/`bus2_voltage`: DC bus voltages
- `grid_voltage_r`/`grid_voltage_s`/`grid_voltage_t`: Grid voltage per phase (R/S/T notation)
- `eps_voltage_r`/`eps_voltage_s`/`eps_voltage_t`: EPS voltage per phase

**Frequency Sensors:**
- `grid_frequency`: Grid frequency
- `eps_frequency`: EPS output frequency

## Why This Change

The `coordinator_mixins.py` already had property mappings for these sensors but they were missing from `SENSOR_TYPES`, causing them to fail silently. Users with 12000XP and other single inverter setups were missing these metrics.

## Test plan

- [x] All 117 tests pass
- [x] ruff check passes
- [x] mypy strict mode passes
- [x] Silver/Gold/Platinum tier validations pass
- [ ] Test on 12000XP device (user validation needed)

Closes #48

Credit: @garjarbinks for identifying the missing sensors

### DIFF
--- a/.github/workflows/quality-validation.yml
+++ b/.github/workflows/quality-validation.yml
@@ -717,35 +717,37 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
-      - name: Set up Python
-        uses: actions/setup-python@v5
-        with:
-          python-version: '3.12'
-
       - name: Verify websession injection
         run: |
-          echo "Checking API client for websession injection support..."
+          echo "Checking websession injection support..."
 
-          # Check API client __init__ signature
-          if ! grep -q "session:\s*Optional\[aiohttp.ClientSession\]\s*=\s*None" custom_components/eg4_web_monitor/eg4_inverter_api/client.py; then
-            echo "❌ API client does not support session injection"
-            exit 1
-          fi
-          echo "✅ API client supports session parameter"
-
-          # Check coordinator uses injection
+          # Check coordinator uses aiohttp session injection via async_get_clientsession
           if ! grep -q "async_get_clientsession" custom_components/eg4_web_monitor/coordinator.py; then
             echo "❌ Coordinator does not inject session"
             exit 1
           fi
           echo "✅ Coordinator injects Home Assistant session"
 
-          # Check config flow uses injection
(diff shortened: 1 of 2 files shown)


garjarbinks:
I also added a synthetic sensor for total load being eps_power +
consumption_power which feels a better representation in the power flow
chart

-----------------------------------------------
Garreth Jeremiah
Senior Technical Account Manager
Prelude Security
+1.416.500.1844

PR #99, opened 2026-01-26
Update README.md with correct Serial Port Settings

Yippy:
There seems to be some confusion why the Serial Port is not set as 19200


btli:
Cherry-picked to main in b4e55f1. Thanks for the fix!

PR #241, opened 2026-05-31
feat: BMS charge/discharge/force-charge battery-bank sensors (#232)

### LINKED ISSUE #232: Add  binary_sensors for battery bms status fields
### Problem or Use Case

BMS can send some data to inverter indicating when can be charged, discharged and request a forced charge.

The information is available on API and include in model of pylxpweb but not available as HA entities, and not decoded when using registers

### Proposed Solution

add a sensor with information from bmsCharge, bmsDischarge, bmsForceCharge API field and get bms battery status on local connections

https://github.com/joyfulhouse/pylxpweb/blob/77958b759fe017cac529ad0c265c6bce4bc0b0ed/docs/api/LUXPOWER_API.md?plain=1#L421

https://github.com/joyfulhouse/pylxpweb/blob/77958b759fe017cac529ad0c265c6bce4bc0b0ed/src/pylxpweb/models.py#L510

for register 95 on my LXP-lB-BR I have confirmed by using the other integration for some months that they are defined as bitmap, and not as static map as defined on pylxpweb. These values should be mapped for the equivalent fields from cloud data (bmsCharge, bmsDischarge, bmsForceCharge)
https://github.com/joyfulhouse/pylxpweb/blob/77958b759fe017cac529ad0c265c6bce4bc0b0ed/src/pylxpweb/constants/registers.py#L321

0x01 - allow charge / bmsCharge
0x02 - allow discharge / bmsDischarge
0x20 - charge request / bmsForceCharge

### Alternatives Considered

Not find any other information that give this information on this integration.

force Charge/Charge Request - indicates that battery need calibration, it is very useful for who like me keeps battery normally charging to 90% and only sometimes do a full charging to calibrate (limit charge to 101%)

allow Charge - is disabled when battery is full

allow Discharge - never seen it disabled here, probably will be disabled when battery is almost discharged and BMS still alive

### Affected Device Model(s)

All Models

btli:
## Summary

Resolves #232 — surfaces the BMS permission/request flags as Home Assistant entities for the first time. Three new **battery-bank** diagnostic sensors, available in **cloud, local, and hybrid**:

| Sensor | States | Meaning |
|---|---|---|
| **BMS Charge Allowed** | Allowed / Blocked | BMS permits charging (Blocked when full) |
| **BMS Discharge Allowed** | Allowed / Blocked | BMS permits discharging (Blocked when empty) |
| **BMS Force Charge Request** | Requested / Idle | BMS requesting a full calibration charge (read-only) |

## Root cause / key finding

The issue author reported that input **register 95 is a bitmap**, while `pylxpweb` modeled it as an enum (`Idle/StandBy/Active`). Investigation confirmed the bitmap: the enum values are exactly the bit combinations of `0x01`(charge)|`0x02`(discharge), and `0x01` was even labelled *"Unknown(1)"* — i.e. "charge-allowed, discharge-blocked" that had never been decoded. The cloud `bmsCharge`/`bmsDischarge`/`bmsForceCharge` booleans are the same three flags, server-decoded.

## What the change does

**pylxpweb** (released as [v0.9.32](https://github.com/joyfulhouse/pylxpweb/releases/tag/v0.9.32)):
- `decode_bms_permissions()` + `BMS_PERMISSION_*` masks (`0x01`/`0x02`/`0x20`)
- reg-95 decode onto `InverterRuntimeData` + `BatteryBankData`; dual-source `BaseInverter.bms_allow_charge/_discharge/_force_charge`; cloud `BatteryBank` delegates to its parent inverter

**Integration** (this PR):
- 3 enum `SENSOR_TYPES` entries + `BATTERY_BANK_CORE_KEYS`, wired through the unified `build_battery_bank_sensors` adapter (cloud + local converge on the same keys)
- `manifest.json` / `requirements-test.txt` → `pylxpweb>=0.9.32`; `DATA_MAPPING.md` documented
- `Force Charge Request` is deliberately named to not collide with the writable "Forced Charge" control (holding reg 21 bit 11)

## Test plan

- **pylxpweb:** 2067 tests pass, `mypy --strict` clean, ruff clean (incl. new `test_bms_permissions.py` covering every bit combination + dual-source paths)
- **Integration:** 854 tests pass, mypy clean, ruff clean (new parity tests in `test_battery_bank_adapter.py`, `test_mode_parity.py`)
- **Live confirmation:** read reg 95 over Modbus from both hybrid inverters and compared to the cloud booleans — 18kPV `4512670118` and FlexBOSS21 `52842P0581` both `reg95=0x03` → `(charge, discharge, force)=(True, True, False)`, **exactly matching** `bmsCharge`/`bmsDischarge`/`bmsForceCharge`

### DIFF
--- a/.beads/interactions.jsonl
+++ b/.beads/interactions.jsonl
@@ -1,0 +1,4 @@
+{"id":"int-9da94f4a","kind":"field_change","created_at":"2026-04-06T18:37:59.240053Z","actor":"bryanli","issue_id":"eg4-1775497212644-112-388d38fb","extra":{"field":"status","new_value":"closed","old_value":"open","reason":"Closed"}}
+{"id":"int-e5750bdc","kind":"field_change","created_at":"2026-04-06T18:42:16.811001Z","actor":"bryanli","issue_id":"eg4-1775497213366-118-70ff7fa9","extra":{"field":"status","new_value":"closed","old_value":"open","reason":"Closed"}}
+{"id":"int-06227459","kind":"field_change","created_at":"2026-04-06T18:47:43.22275Z","actor":"bryanli","issue_id":"eg4-1775497213601-120-11784d2b","extra":{"field":"status","new_value":"closed","old_value":"open","reason":"Closed"}}
+{"id":"int-57468201","kind":"field_change","created_at":"2026-04-06T18:47:53.304813Z","actor":"bryanli","issue_id":"eg4-1775497214254-125-cdce411f","extra":{"field":"status","new_value":"closed","old_value":"open","reason":"Closed"}}
--- a/.claude/commands/sprint-loop.md
+++ b/.claude/commands/sprint-loop.md
@@ -1,0 +1,101 @@
+---
+description: "Continuous sprint orchestrator: triage, plan, execute, review, merge in a loop. Pass a number to limit iterations (default: 1)."
+---
+
+# Sprint Loop
+
+Orchestrate continuous sprint cycles. Each iteration runs a full sprint:
+triage -> plan -> execute -> review -> merge -> retro.
+
+## Input
+
+- `$ARGUMENTS` — number of sprint iterations to run (default: 1, "all" for continuous until backlog empty)
+
+## Workflow
+
+### Initialization
+
+1. Run bd doctor to ensure beads is healthy:
+   ```bash
+   bd doctor
+   ```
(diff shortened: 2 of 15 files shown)


btli:
Superseded by #242, which is the same change re-based on the `3.3.0-beta` branch (so the beta tag `v3.3.0-beta.1` and the PR share one branch). Closing in favor of #242.

Issue #153, opened 2026-02-13
v3.2.0-beta.30 - Battery Backup Mode Failure


trinitron79:
Looks like it might not be possible to enable Battery Backup Mode via local calls.

<img width="737" height="112" alt="Image" src="https://github.com/user-attachments/assets/a2476c3c-bf2e-4c18-b1fe-ba3280f1fffa" />

HA Logs:
2026-02-13 14:29:32.965 ERROR (MainThread) [homeassistant.components.websocket_api.http.connection] [140362007284768] Error during service call to switch.turn_on: Working mode FUNC_BATTERY_BACKUP_CTRL not available via local transport

I was talking to Sig Solar about this setting when my account was in API Lockout.  This mode is NOT something that you can find in the app when logging in locally via the EG4 Phone App, nor is it something on the screen you can select.  So possible not something you can fix, but maybe still there just not exposed locally?

btli:
## Fix: LOCAL mode support for Battery Backup and Grid Peak Shaving switches

**Root cause**: Holding registers 179 and 233 were not mapped in pylxpweb's `REGISTER_TO_PARAM_KEYS`, so these functions had no local Modbus write path. The switches were created but toggling them raised "not available via local transport".

**Investigation**: Used a Modbus probe script to read raw register values while toggling functions on/off via the web app:
- **FUNC_BATTERY_BACKUP_CTRL** → holding register 233, bit 1 (confirmed via read + write test)
- **FUNC_GRID_PEAK_SHAVING** → holding register 179, bit 7 (confirmed via read test)

**Fix** (commit e799554):
- Added registers 179 and 233 to `REGISTER_TO_PARAM_KEYS` with confirmed bit positions
- Added both registers to `_read_modbus_parameters()` register ranges (coordinator_local.py)
- Mapped `FUNC_GRID_PEAK_SHAVING` and `FUNC_BATTERY_BACKUP_CTRL` to actual parameter names in `_WORKING_MODE_PARAMETERS` (was `None`)
- Restored both to `modbus_supported_params` in entity creation filter

All 5 working mode switches now have full LOCAL mode parity with Cloud:
| Working Mode | Register | Bit | Status |
|-------------|----------|-----|--------|
| AC Charge | 21 | 7 | Existing |
| PV Charge Priority | 21 | 11 | Existing |
| Forced Discharge | 21 | 10 | Existing |
| Grid Peak Shaving | 179 | 7 | **New** |
[...]

PR #223, opened 2026-04-30
fix: pin total_increasing sensor dips to suppress recorder warnings (#218)

btli:
## Summary

Closes #218.

**Bug**: HA's recorder logged `Entity ... has state class total_increasing, but its state is not strictly increasing` on every poll for `sensor.flexboss21_*_consumption` and `consumption_lifetime` (Cloud/HTTP mode, FlexBOSS21).

**Root cause**: Cloud `consumption_lifetime` is computed as `pv + discharge + grid_import − charge − grid_export`, with each component rounded to 1 decimal place at the API. The result jitters by ±0.1 kWh between polls (e.g. 2917.1 → 2917.0). HA's recorder warns whenever `total_increasing` dips by ≤10% (drops >10% are silently treated as meter resets). Single-decimal precision is fundamental to the API, so the right fix is at the sensor entity layer.

**Fix**: Added `_guard_total_increasing()` in `base_entity.py` that pins downward dips ≤10% to the previous high-water mark, matching HA recorder's `0.9 * old > new` reset condition exactly. Wired into `native_value` on `EG4BaseSensor`, `EG4BaseBatterySensor`, and `EG4BatteryBankEntity` via a per-instance `_last_reported_value` cache.

**Daily resets pass through unchanged.** A `consumption` rolling 14.4 kWh → 0.0 at midnight is a 100% drop, far above 10%; the guard treats it as a reset and lets the 0 through, updating the cache so subsequent comparisons baseline against 0. Same for inverter replacements (lifetime → 0).

## Test plan

- [x] 19 new unit tests in `tests/test_base_entity.py` covering:
  - Issue scenario: `consumption_lifetime` 2917.1 → 2917.0 pinned
  - Daily reset: `consumption` 14.4 → 0.0 passes through
  - Inverter replacement: lifetime 2917.1 → 0.0 passes through
  - Drop just below 10% (10.07%) treated as reset
  - Drop exactly at 10% suppressed (HA recorder's warning boundary)
  - Consecutive dips keep cache pinned
  - Recovery after dip resumes normal progression
  - StrEnum and string state-class normalization
  - None / non-numeric / non-`total_increasing` pass-through
- [x] All 41 base-entity tests pass
- [x] 903 of 908 integration tests pass (5 pre-existing failures in `TestDiscoverSerialDeviceWithErrors` unrelated to this fix)
- [x] `ruff check` clean, `ruff format` clean
- [x] `mypy --strict` clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #218, opened 2026-04-11
FB21 Consumption sensor warning


parsissn:
### Integration Version

v3.3.0-alpha.1

### Home Assistant Version

2026.4.1

### Connection Mode

Cloud (HTTP)

### Device Model(s)

FlexBOSS 21

### Describe the Bug

None critical bug report. This error occurs in my log files so reporting it back for visibility. 

Logger: homeassistant.components.sensor.recorder
Source: components/sensor/recorder.py:431
integration: Sensor (documentation, issues)
First occurred: 9:50:10 AM (2 occurrences)
Last logged: 9:50:10 AM

Entity sensor.flexboss21_52842p0126_consumption from integration eg4_web_monitor has state class total_increasing, but its state is not strictly increasing. Triggered by state 14.2 (previous state: 14.3) with last_updated set to 2026-04-11T13:45:50.920590+00:00. Please create a bug report at https://github.com/joyfulhouse/eg4_web_monitor/issues
Entity sensor.flexboss21_52842p0126_consumption_lifetime from integration eg4_web_monitor has state class total_increasing, but its state is not strictly increasing. Triggered by state 2917.0 (previous state: 2917.1) with last_updated set to 2026-04-11T13:45:50.920765+00:00. Please create a bug report at https://github.com/joyfulhouse/eg4_web_monitor/issues

### Steps to Reproduce

Automatically generated in HA core logs. No manual steps taken to reproduce. 

### Home Assistant Logs

```shell

```

btli:
Hi @parsissn — opened a fix for this in PR #223. Would appreciate your help validating before we merge.

**Branch**: `fix/issue-218-fb21-consumption-sensor-warning`
**Commit**: `8f4b8df270d43db96f054689daa4d00082b4af4c`
**PR**: https://github.com/joyfulhouse/eg4_web_monitor/pull/223

### Root cause

Your `consumption` and `consumption_lifetime` are computed (cloud mode) as `pv + discharge + grid_import − charge − grid_export`, with each component rounded to 1 decimal place at the EG4 API. The result jitters by ±0.1 kWh between polls (e.g. 2917.1 → 2917.0). HA's recorder warns whenever a `total_increasing` sensor dips by ≤10% (drops >10% are treated as silent meter resets).

### What the fix does

[...]

PR #244, opened 2026-06-02
fix: daily consumption resets at midnight in LOCAL mode (#227)

### LINKED ISSUE #227: Daily consumption not resetting to zero
### Integration Version

v3.2.0

### Home Assistant Version

2026.4.4

### Connection Mode

Local (WiFi Dongle)

### Device Model(s)

18kPV

### Describe the Bug

This PR: #40 seems to have been intended to make daily total_increasing sensors get reset to zero at midnight, but for my 18kpv it seems not to be working in the case of the (daily) consumption sensor.

Here are some screenshots of what it looks like in my install:

Image<br />

Image<br />

Image

### Steps to Reproduce

1. go to devices & services -> EG4 -> 18kpv
2. look at the consumption sensor
3. note that it never resets to zero

### Home Assistant Logs

```shell
no relevant logs
```

### Screenshots

See above.

btli:
## Summary

Fixes #227 — the daily **Consumption** sensor never resets to zero at midnight in **LOCAL** mode (WiFi Dongle / Modbus). It holds at the day's peak and only rises when consumption surpasses that peak.

## Root cause

In LOCAL mode, `consumption`/`consumption_lifetime` are *computed* from an energy balance (`_energy_balance(yield, discharging, grid_import, charging, grid_export)`). The local coordinator then ran `_clamp_computed_energy`, an **unbounded monotonic guard**:

```python
if prev is not None and curr is not None and curr < prev:
    sensors[key] = prev   # rejects the midnight 25 kWh -> 0 reset
```

It was added (#153) to smooth tiny multi-register timing jitter, but being unbounded it also rejected the **legitimate midnight reset**, pinning the daily sensor at its peak.

**Why LOCAL-only:** the clamp lives only in `coordinator_local` (MODBUS / DONGLE / LOCAL paths). CLOUD reads the server-computed `todayUsage` directly, and HYBRID never calls the clamp — so cloud and hybrid reset correctly. (The "upgrade from an old config" theory does not apply: the clamp is unconditional, independent of `data_validation`.)

## Fix

`consumption` is `state_class: total_increasing`; Home Assistant detects meter resets (>10% drops) natively and absorbs sub-10% dips. So the integration must not clamp. This removes:

- `_clamp_computed_energy` method
- `_COMPUTED_ENERGY_KEYS` constant
- both call sites (per-inverter + parallel group)

Tiny same-day derivation jitter can now reach HA, which treats it as noise — strictly better than permanently pinning the sensor at its peak.

## Test plan

- ❌→✅ New regression `test_midnight_reset_consumption_not_clamped` (prev peak 50 → current 0 must publish 0) — **fails** with the clamp present, **passes** after removal.
- Removed obsolete `TestComputedEnergyClamp` (it asserted the buggy held-at-peak behavior).
- ✅ Full suite: **853 passed**
- ✅ `ruff check` + `ruff format` clean
- ✅ `mypy --strict` clean
- ✅ Platinum + Gold tier validators pass

### DIFF
--- a/custom_components/eg4_web_monitor/coordinator_local.py
+++ b/custom_components/eg4_web_monitor/coordinator_local.py
@@ -73,11 +73,6 @@ if TYPE_CHECKING:
 # concrete local types lets the typed seam verify the split_phase write.
 _LOCAL_REGISTER_TRANSPORTS = (ModbusTransport, ModbusSerialTransport, DongleTransport)
 
-# Derived from energy balance arithmetic; small decreases are normal
-# timing artifacts (not corruption).  Clamped to prevent HA
-# total_increasing warnings.
-_COMPUTED_ENERGY_KEYS = frozenset({"consumption", "consumption_lifetime"})
-
 # Minimum battery serial length to consider valid.  Shorter serials are
 # likely truncated register reads from incomplete CAN bus transfers and
 # are skipped to avoid creating phantom battery entities.
@@ -590,39 +585,6 @@ class LocalTransportMixin(_MixinBase):
             model="Parallel Group",
         )
 
-    def _clamp_computed_energy(
-        self,
-        device_id: str,
-        sensors: dict[str, Any],
-    ) -> None:
-        """Clamp computed energy values so they never decrease.
-
-        Small decreases are normal timing artifacts from reading multiple
-        registers at different instants.  Clamping prevents HA
-        total_increasing state class warnings.
-        """
-        if not self.data:
-            return
-        prev_sensors = (
-            self.data.get("devices", {}).get(device_id, {}).get("sensors", {})
-        )
-        if not prev_sensors:
-            return
-
-        for key in _COMPUTED_ENERGY_KEYS:
-            prev = prev_sensors.get(key)
-            curr = sensors.get(key)
-            if prev is not None and curr is not None and curr < prev:
-                _LOGGER.debug(
-                    "Clamping %s for %s: %.1f -> %.1f (keeping %.1f)",
-                    key,
-                    device_id,
-                    prev,
-                    curr,
-                    prev,
-                )
-                sensors[key] = prev
-
     async def _process_single_local_device(
         self,
         config: dict[str, Any],
@@ -1046,8 +1008,6 @@ class LocalTransportMixin(_MixinBase):
                     sensors.get("rectifier_power"),
                     sensors.get("grid_import_power"),
                 )
-
-                self._clamp_computed_energy(serial, sensors)
 
                 processed["devices"][serial] = device_data
                 device_availability[serial] = True
@@ -1796,9 +1756,7 @@ class LocalTransportMixin(_MixinBase):
                                 group_sensors[vkey] = val
                         break
 
-            # Clamp computed energy values before storing the parallel group.
             group_device_id = f"parallel_group_{group_name.lower()}"
-            self._clamp_computed_energy(group_device_id, group_sensors)
 
             group_sensors["parallel_group_last_polled"] = dt_util.utcnow()
 
--- a/tests/test_coordinator.py
+++ b/tests/test_coordinator.py
@@ -2565,6 +2565,44 @@ class TestParallelGroupAggregationMath:
         assert total_consumption == 7600  # 2800 + 4800
 
     @pytest.mark.asyncio
+    async def test_midnight_reset_consumption_not_clamped(self, hass, local_entry):
+        """Regression #227: daily consumption must reset to zero at midnight.
+
+        The removed ``_clamp_computed_energy`` held computed consumption at its
+        daily peak and rejected any decrease, so the midnight drop-to-zero never
+        reached Home Assistant.  HA's ``total_increasing`` state class detects
+        meter resets natively, so the integration must publish the decrease as-is
+        instead of pinning the sensor to the previous (peak) value.
+        """
+        coordinator = EG4DataUpdateCoordinator(hass, local_entry)
+        # Previous cycle = yesterday's accumulated peak.
+        coordinator.data = {
+            "devices": {
+                "parallel_group_a": {
+                    "sensors": {"consumption": 50.0, "consumption_lifetime": 5000.0}
+                },
+                "MASTER001": {
+                    "sensors": {"consumption": 25.0, "consumption_lifetime": 2500.0}
+                },
+                "SLAVE002": {
+                    "sensors": {"consumption": 25.0, "consumption_lifetime": 2500.0}
+                },
+            }
+        }
+        # Midnight: per-inverter daily consumption resets to 0; lifetime climbs.
+        processed = self._make_processed(
+            {"consumption": 0.0, "consumption_lifetime": 2500.1},
+            {"consumption": 0.0, "consumption_lifetime": 2500.1},
+        )
+        await coordinator._process_local_parallel_groups(processed)
+
+        sensors = processed["devices"]["parallel_group_a"]["sensors"]
+        # Daily consumption reflects the reset — NOT held at the previous peak.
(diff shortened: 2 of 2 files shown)


btli:
### Code review

Found 1 issue:

1. Removing the unbounded `_clamp_computed_energy` correctly fixes the #227 midnight-reset freeze, but on `main` it reintroduces the #218 jitter warnings. The clamp (added in af6aca2 / #153) was also suppressing HA `total_increasing` "state is not strictly increasing" warnings: `consumption`/`consumption_lifetime` are computed from a 5-component energy balance, each input rounded to 0.1 kWh, so the sum can dip ~0.1 kWh between polls even when load is unchanged. That is exactly what #218 reported.

   The proper, reset-safe fix is the **bounded** entity-layer guard `_guard_total_increasing` in open PR #223 (suppresses sub-threshold dips, lets the >threshold midnight reset through). It is not yet on `main`. Recommend landing #223 together with (or before) this PR — or rebasing this branch on top of it — so #227 is fixed without regressing #218. This PR alone is correct for #227; the note is about merge sequencing.

   consumption_lifetime sensor (`total_increasing`): https://github.com/joyfulhouse/eg4_web_monitor/blob/aef2a4fc1ebde23970df6e946185641d35b53bb7/custom_components/eg4_web_monitor/const/sensors/inverter.py#L692-L698

ivanfmartinez:
@btli the modbus docs indicate that luxpower inverter have today energy values on Input Registers. 
Other inverters / EG4 does not have this registers ?

Screenshot_2026-06-02_11-53-05

ivanfmartinez:
Here all that I found on luxpower doc about today energy

Screenshot_2026-06-02_11-56-07


Screenshot_2026-06-02_11-56-45

Screenshot_2026-06-02_11-57-16


Screenshot_2026-06-02_11-57-56

btli:
The registers exist but in parallel group setting it doesn't seem to provide accurate data. Investigating.

btli:
@ivanfmartinez thanks for the nudge on the load registers — it sent us down a useful rabbit hole, and I'd value your take before we commit to a direction.

**What we confirmed on live hardware** (parallel pair: FlexBOSS21 master `…0581` + 18kPV slave `…0118`, both Modbus TCP, plus a GridBOSS), cross-checked against the cloud:

| Quantity | slave | master | group |
|---|---|---|---|
| **reg 171 `Eload_day`** (local) | 20.1 | **0.0** | — |
| Cloud **per-inverter** `todayUsage` | 20.1 | **0.0** | — |
| Cloud **parallel-group** `todayUsage` (`getInverterEnergyInfoParallel`) | — | — | **40.5** |
| `energy_balance` (`pv+dis+imp−chg−exp`) | 19.1 | 19.2 | ~38–41 |

Two findings:

1. **reg 171 (Eload) matches the cloud's *per-inverter* `todayUsage` exactly** — including the master reading **0.0** while it imports ~20 kWh (it passes grid through to the parallel bus; it directly serves no backup load). So the register is reliable; it measures *inverter-served load*.

2. **The cloud's *group* `todayUsage` is 40.5**, which matches `energy_balance`, **not** the sum of per-inverter Eload (20.1). The ~20 kWh gap is grid-direct / GridBOSS-Load-port load that never flows through an inverter's Eload.

So per-inverter Eload and whole-home consumption are **different scopes**, and the cloud reports them as two separate numbers that don't sum. Our single `consumption` sensor currently uses `energy_balance` so it sums toward the household — at the cost of not matching the cloud's per-unit value (master 19 vs cloud 0).

**Where we're unsure — your take?**
- Should per-inverter `consumption` mirror the cloud's per-unit value (**reg 171 / Eload**, master = 0), or stay an `energy_balance` household-share estimate (master ≈ 19)?
- Or is the right model **two sensors** — a per-inverter *"load energy"* (reg 171, raw) plus a site/household *"consumption"* (group `energy_balance` / GridBOSS UPS+Load CTs) — rather than overloading one sensor?
- For your inverters, does reg 171 (`Eload_day`/`Eload_all`, regs 171–173) track real consumption, and do you have grid-direct loads that bypass it?

For reference, the #227 fix in this PR (removing the monotonic clamp so the daily sensor resets) is correct and independent of all the above; this question is only about *which value* the consumption sensor should carry. Test build with the clamp fix is `v3.3.0-beta.5`.

ivanfmartinez:
In my case the most of registers indicates only the per inverter data as you find.
I never liked the combined value on luxpower site, got some discrepancies and left my inverters as separate.

This 171 register specifically in my case gives same value on both inverters.
Screenshot_2026-06-02_13-22-02

I had an external meter (broken and I will replace) and the luxpower apparently have some issues on the external CT measurements. In my case as split-phase when using multiple phases one phase can be injecting and other getting energy from grid. The power sensors are mutually exclusive since I installed my devices. It indicates you are Injecting or getting from grid.

In my case as I use two split-phase to make 3-phase 127/220V these energy and power consumption from luxpower are not trustable.

In more than 6 months I never saw both import/export power with values, even when my other meter indicate that one phase is injecting and other getting from grid. And probably the energy is computed using this metered powers which will give wrong values on energy when phases are not balanced. I reported this to luxpower and they not solved the question.
I think that they transformed the single phase inverter on two-phase but kept some of constraints that are only valid for single-phase.

"Or is the right model two sensors — a per-inverter "load energy" (reg 171, raw) plus a site/household "consumption" (group energy_balance / GridBOSS UPS+Load CTs) — rather than overloading one sensor?"

Personally I like having separate sensors, having per inverter sensors allow to easily check the data instead of using a single combined sensor.

There are some conditions that can change on firmware updates, I remember that some values are not given on firmware 24 and started on firmware 25 (and no information about this changes on Changelog)

I think that for most cases keeping data as the device gives is best, combining on integration will maybe work for one firmware version but not for others.

btli:
@ivanfmartinez exactly the validation we needed — thank you. Your "I like having separate sensors ... keeping data as the device gives is best; combining on integration will maybe work for one firmware version but not for others" is the conclusion we landed on, and it's shipped.

**[`v3.3.0-beta.6`](https://github.com/joyfulhouse/eg4_web_monitor/releases/tag/v3.3.0-beta.6)** adds per-inverter **Load Energy** / **Load Energy (Lifetime)**, read **verbatim** from `Eload` regs 171/172 — no computation, no combining. They equal the cloud's per-inverter `todayUsage`/`totalUsage` to the decimal in every mode (validated live, including a parallel master reading `0` while importing ~20 kWh). That raw passthrough is exactly what's robust to the fw-24-vs-25 register variability you flagged — we surface what the register holds and let HA handle it.

Whole-home **Consumption** is left unchanged (energy balance / GridBOSS CT / cloud group), so this is **non-breaking**: anyone using it in the Energy Dashboard keeps working, and you now also get the clean per-inverter raw value to inspect.

Good detail that reg 171 reads the **same on both** your inverters — that's your 2×split-phase→3-phase 127/220V topology (the inverters mirror the combined load); in a parallel group the master reads 0 instead. Surfacing the raw per-inverter register is the right call either way, which is why we didn't overload one sensor. Your untrustworthy-CT observation matches ours too: `energy_balance` is fine for *daily* but unreliable for *lifetime* (the per-inverter lifetime components wrap), so whole-home lifetime comes from the cloud group / GridBOSS CT, never per-inverter balance.

The #227 clamp fix in this PR is unchanged and still the thing to test here — Load Energy just rode into the same beta. Full write-up: [DATA_MAPPING.md → "Consumption vs Load Energy"](https://github.com/joyfulhouse/eg4_web_monitor/blob/3.3.0-beta/docs/DATA_MAPPING.md).

btli:
Closing: branch content is fully patch-equivalent to main — the midnight-reset fix (#227) shipped in 3.3.0-beta.5 / 3.3.0.

PR #247, opened 2026-06-05
fix: PV Charge Power targets reg 74 in kW (fixes set-1-reads-0 bounce)

btli:
## Problem
Setting `number.<model>_<serial>_pv_charge_power` on a Modbus/hybrid-connected inverter (e.g. 18kPV) did not stick — setting 1 kW read back 0. Cloud-connected inverters (e.g. FlexBOSS) worked.

## Root cause
Same register on both inverters, but **different transports → different code paths**. The local/hybrid path wrote register **64** (`HOLD_CHG_POWER_PERCENT_CMD`, a 0-100% limit) with a lossy `int(kW/15*100)` conversion and read it back with `int(%/100*15)` → set 1 kW wrote 6%, read `int(0.9)=0`. The cloud path already used register **74** correctly, so cloud inverters were unaffected.

## Fix
Retarget `PVChargePowerNumber` to register **74** (`HOLD_FORCED_CHG_POWER_CMD`) in kW, mirroring the proven `ACChargePowerNumber` (reg 66, also 100W units): read raw÷10, write kW×10. Cloud path unchanged. Both transports now agree → no bounce.

Hardware-verified against live inverters: FlexBOSS reg74=20→2.0 kW, 18kPV reg74=120→12.0 kW.

## Dependency
Requires `pylxpweb>=0.9.35b7` (joyfulhouse/pylxpweb#174) which adds reg 74 to the local register map. Manifest pin bumped.

## Tests
New regressions: no-bounce (raw 10→1 kW), zero-value, and hybrid out-of-range fallback (guards against double-scaling). Full suite green (862 passed).


btli:
Superseded by the v3.3.0 release — the PV Charge Power reg-74 fix is now on `main` via the 3.3.0-beta merge (commit 529104d). Closing as redundant.

Issue #207, opened 2026-03-30
Add support for setting Forced Discharge Power Rate (%) via local (dongle/IP) connection


DevTodd:
### Problem or Use Case

First of all, thanks for all the hard work!

I currently have multiple buildings on the property and my grid sell back is actually feeding the main house panel.  I currently use private code based on Lux Power inverter data to set the forced discharge power to regulate the power being fed back to the "grid" which is the house.  I use CTs on the main panel, feed that to HA and adjust the discharge power to match the house power.

The integration currently supports enabling/disabling forced discharge, but does not expose control of the Forced Discharge Power Rate (%).

This parameter is available on the inverter and can be written locally via the dongle (TCP/IP), without requiring cloud access.

### Proposed Solution

From testing and comparison with another working integration:

The inverter supports Forced Discharge Power Rate (%)
This is a writable holding register:
Register: 82
Range: 0–100 (%)
Forced discharge itself is already exposed in this integration as a switch (register 21 bitmask), but the rate control is missing

In another implementation, this is exposed as:

```python
{
  "name": "Forced Discharge Power Rate (%)",
  "register_address": 82,
  "min_val": 0,
  "max_val": 100
}
```

So the functionality is confirmed working at the protocol level.

Thank you for taking the time to look this over and consider it!

### Affected Device Model(s)

18kPV

btli:
A fix for this issue is now available in [v3.3.0-alpha.1](https://github.com/joyfulhouse/eg4_web_monitor/releases/tag/v3.3.0-alpha.1).

**To install:**
1. Go to HACS → Integrations → EG4 Web Monitor
2. Click the 3-dot menu → Redownload → Show beta versions → Select **v3.3.0-alpha.1**
3. Restart Home Assistant

Please test and report any issues. This is a pre-release for community testing — stable release will follow after validation.

icepop456:
I tried the update version on my Flexboss21.  I get an error message trying to update the power rate %

<img width="534" height="1104" alt="Image" src="https://github.com/user-attachments/assets/8e1a2f0e-fecb-403d-8e95-58360fe5987f" />



**New Issue:** 
I also noticed that the AC Power level toggles between 0 and setpoint (5 kW in my case).  This did not happen on stable 3.2

<img width="524" height="839" alt="Image" src="https://github.com/user-attachments/assets/d6359059-2a20-49d2-bf48-e29d6d4091ab" />

DevTodd:
[...]

Issue #201, opened 2026-03-21
WiFi dongle drops TCP connection during parameter writes (LOCAL-only mode broken)


btli:
### Description

WiFi dongle transport drops the TCP connection during parameter write sequences (e.g., toggling Off Grid Mode, PV Charge First), causing `Socket error: Connection lost`. Reads work fine; only writes are affected.

### Root Cause

Parameter writes require a multi-step **read-modify-write** cycle on holding registers:
1. Read current register value (function code 0x03)
2. Modify the bit field locally
3. Write the new value (function code 0x10/0x06)

The dongle connection drops during this multi-transaction sequence, likely due to firmware timeout or cloud server connection priority.

### Impact

- **HYBRID mode**: Works — `_execute_local_with_fallback()` correctly falls back to cloud API
- **LOCAL-only mode**: All parameter writes fail with no fallback available
- Affects all switch controls: Off Grid Mode, PV Charge First, Forced Charge, etc.

### Reproduction

1. Configure integration with WiFi dongle only (LOCAL mode)
2. Toggle any switch control (e.g., Off Grid Mode)
3. Observe: `Failed to write parameter FUNC_GREEN_EN: Socket error: Connection lost`

### Logs

```
Logger: pylxpweb.transports.dongle
Source: runner.py:289
Socket error communicating with dongle: Connection lost

Logger: custom_components.eg4_web_monitor.coordinator
Source: coordinator.py:629
Failed to write parameter FUNC_GREEN_EN: Socket error: Connection lost
```

### Proposed Fix

Add retry/reconnect logic to pylxpweb's dongle transport `write_named_parameters()` path:
- Reconnect on socket error and retry the write sequence
- Consider a short delay between read and write steps to reduce connection pressure
- May need to be addressed in pylxpweb rather than the integration

### Environment

- FlexBOSS 21 firmware: FAAB-2525
- GridBOSS firmware: IAAB-1600
- Connection: WiFi dongle (LOCAL-only)
- Integration: v3.2.0

### Related

- Discovered during #194 investigation
- Cloud API fallback works correctly (confirmed by reporter)

### Labels

bug

btli:
[...]

PR #249, opened 2026-06-10
fix: add local forced discharge read/write support

DevTodd:
Add local TCP/IP support for forced discharge settings using registers 82 and 83.

This fixes forced discharge values not being read or refreshed correctly when using the WiFi dongle transport. The integration now reads both forced discharge registers with read_parameters(82, 2), updates the parameter cache, and keeps the Home Assistant number entities in sync when values are changed on the inverter or through Home Assistant.

Tested on an EG4 Hybrid inverter using the local WiFi dongle transport.

Verified:
- Forced Discharge Power Rate reads correctly at startup
- Forced Discharge SOC Limit reads correctly at startup
- Values update correctly when changed through Home Assistant
- Values refresh correctly when changed directly on the inverter
- Integration loads without circular import errors

Refs joyfulhouse/eg4_web_monitor#207


btli (changes requested):
Thanks for this @DevTodd — local forced-discharge control is wanted (#207), and the register verification is the genuinely hard part: regs **82/83 match our canonical register map exactly** (`HOLD_FORCED_DISCHG_POWER_CMD` / `HOLD_FORCED_DISCHG_SOC_LIMIT`), so your hardware testing lines up with what we have. Requesting changes on the mechanics, with a concrete path below.

**Blockers**

1. **The branch is missing a file and cannot load.** `const/__init__.py` imports `FORCED_DISCHARGE_POWER_MIN/...` from `const/limits.py`, but `limits.py` is unchanged in this PR — the integration fails at import:
   ```
   ImportError: cannot import name 'FORCED_DISCHARGE_POWER_MIN' from 'custom_components.eg4_web_monitor.const.limits'
   ```
   Looks like an uncommitted file on your working tree.

2. **Wrong base branch.** This targets `main` (v3.3.0 stable). Active development happens on the **`3.4.0-beta`** line, where `coordinator_local.py` and `number.py` have changed substantially (v3.4.0-beta.4 shipped today) — this diff won't apply there. Please rebase onto `3.4.0-beta`.

**Design (this is the big one, and it makes the PR ~85% smaller)**

3. The ~125-line probe cascade in `coordinator_local.py` (`_try_call`/`_try_read` over `transport`/`client`/`_client`/…, `_get_raw_value` over dict/list/attr shapes) is guessing at the pylxpweb API. The supported path: pylxpweb's parameter refresh already populates the coordinator's parameter cache keyed by `HOLD_*` names — registers 82/83 just need to be added to pylxpweb's register→param-key map, which is a 2-line library change **we'll handle on our side**. After that, the integration side is just two ~30-line parameter-backed number entities following the existing `AC Charge Power` / `AC Charge SOC Limit` pattern — no custom register reads at all.

4. **Per-entity `async_update` doing its own reads** bypasses the coordinator's scheduling. On the WiFi dongle there is a single TCP slot — ad-hoc entity reads contend with coordinator polls, which is exactly the failure mode we spent #201/#226 hardening against. Number entities should read from the coordinator's parameter cache and write through the inverter's parameter-write API (which already has retry/ack handling).

**Needs adjudication**

5. Your docstring says reg 82 is "**tenths of a kW**"; our canonical table and the cloud parameter it mirrors say **percent (0–100)**. This is the same units conflation that bit us on PV Charge Power (reg 64 % vs reg 74 in 100 W units). Could you check on your unit: set Forced Discharge Power to a known value on the inverter panel (say 50%) and tell us the raw register value you read back? That settles it conclusively.

**Smaller items**

6. No tests (CI gates: full pytest + mypy strict + ruff), three identical commit messages (please squash), and new entities need `strings.json` + all 14 locale files per repo convention.

**Two ways forward — your pick:**
- **(a)** You rebase onto `3.4.0-beta` and slim it to the param-cache pattern once we land the pylxpweb mapping (we'll ping you here when it's on the staging branch), or
- **(b)** we port the feature onto the beta line ourselves and credit you (`Co-authored-by` + tested-on-hardware credit) — your register verification is the part that mattered.

Either way this lands. Thanks again for digging into the registers on real hardware.

btli:
Following up on the review: rather than another round-trip, I've taken **path (b)** — the feature is implemented on the development branches with you credited as **co-author on both commits** (integration + pylxpweb), since your register verification was the part that mattered. It ships in the next release; details on #207 (reopened to track it).

What landed, relative to this PR:
- Registers 82/83 confirmed and pinned in the register-contract harness — your hardware verification holds.
- The transport plumbing went through pylxpweb's parameter map + cloud setters instead of the probe cascade, which shrank the integration side to two ~50-line entities on the existing parameter-cache pattern (and keeps the dongle's single TCP slot uncontended).
- **Units**: implemented as percent (0–100) per the canonical table, the cloud parameter naming, and #207's own title. If your set-50-read-raw check ever shows otherwise, that's a one-line scale fix — still interested in the answer when you get a chance.

Closing this as superseded by the co-authored port. Thanks again for the hardware work — exactly the kind of contribution that's hard to do without the gear on the bench.

DevTodd:
Thanks for fixing those issues.  I just sat down to tackle the request, but even better than you took care of it.  As for the adjudication; The measurement for force discharge power is in kW, not percent.  I discovered that when I enter a value on the inverter such as 2.5kW it returned 25 on the raw register value.  For the forced discharge stop percent it was returning the same value 18 = 18, etc.  See the images below:

image
image

btli:
Confirmed — you were right, and the fix is in. Your set-X-read-raw legwork (2.5 kW → raw 25) settles it: register 82 uses the same 100 W-unit encoding as the forced-charge power register, not percent. Your original docstring's "tenths of a kW" was correct all along; the percent reference it was checked against (the 18KPV Modbus PDF) is what's wrong for this register. The cloud maintain page agrees — its field is "Forced Discharge Power 1(kW)", float, range [0, 25.5] (the raw uint8 ceiling, 255 × 100 W).

What changed on the development branches (with you as co-author on these commits as well):
- The entity is now **Forced Discharge Power** in **kW** (0–25.5, 0.1 steps): local writes convert kW → raw ×10, cloud writes send float kW directly. Stop SOC stays percent, exactly as your readings showed (18 = 18).
- Live-verified today on an 18kPV and a FlexBOSS21 via cloud write → readback → revert: 12 → 2.5 → 12 kW round-trips cleanly on both, and fractional kW is accepted.
- Bonus from the same maintain page: the third field, *Stop Discharge Volt 1(V)*, was located at **register 202** (`_12K_HOLD_STOP_DISCHG_VOLT`) by single-register window reads and confirmed by write/readback/revert on both units — the voltage-regime counterpart of the stop SOC. An entity for it is queued separately.

Thanks again — this is the second time your bench data has directly corrected the register documentation.

Issue #256, opened 2026-06-15
Offline inverter (lost=true) blacks out ALL its entities — partial cloud payload fails pylxpweb model validation


btli:
## Summary

When an inverter is **offline** (cloud `lost: true`), **all** of its Home Assistant entities go `unavailable` — including the `Status` sensor — instead of the device simply reporting "offline". A second, online inverter in the same station (e.g. a FlexBOSS21) is unaffected, which makes it look like the offline unit "broke" on a recent update.

Reported on 3.4.0-beta.13 with an **18kPV** (offline) alongside a **FlexBOSS21** (online): the 18kPV's `Status` and runtime entities were all `unavailable`; the FlexBOSS21 showed normally.

## Root cause

The EG4 cloud returns a **partial** `getInverterRuntime` / `getBatteryInfo` payload for an offline device — it omits the live/aggregate measurement fields but keeps last‑known/config fields (`statusText: "offline"`, per‑string PV, voltages, temps).

`pylxpweb`'s `InverterRuntime` and `BatteryInfo` Pydantic models declared those live fields **required**, so the partial payload failed validation outright:

```
Failed to fetch runtime data for 4512670118: 11 validation errors for InverterRuntime
Unexpected error fetching battery data for 4512670118: 5 validation errors for BatteryInfo
```

`InverterRuntime` missing (required) fields: `ppv, status, pCharge, pDisCharge, batPower, batteryColor, soc, vBat, pinv, prec, peps`
`BatteryInfo` missing (required) fields: `batStatus, soc, vBat, pCharge, pDisCharge`

Because validation raised, `pylxpweb` discarded the **entire** response (including the fields that *were* present). The chain:

1. offline inverter → cloud returns partial runtime (no `status`, `ppv`, battery fields…)
2. `InverterRuntime.model_validate()` raises → `_fetch_runtime_http()` swallows it → `_runtime` stays `None`
3. `BaseInverter.has_data` → `False` (cloud has no transport runtime)
4. integration's `_process_inverter_object()` hits `if not inverter.has_data:` and returns **only diagnostic sensors** — no runtime sensors
5. every runtime entity (incl. `Status`) is `unavailable`

Any inverter would black out identically when offline; the FlexBOSS21 just happened to be online (`lost: false`, full payload).

### Evidence (live)

| field | 18kPV (offline) | FlexBOSS21 (online) |
|---|---|---|
| `lost` | `true` | `false` |
[...]

Issue #82, opened 2026-01-14
Add Entity for Online State


iamcheek:
I noticed this morning in the monitor app that my inverter was offline because my dongle had lost connection. Because of that, I’m missing data for about a 6 hour period. 

I’m not sure which API you’re using, but the public facing reference lists this endpoint
/api/device/v3/summary/currentState/{serialNumber}

One of the properties in the response body is “online” which has a Boolean value. 

Having an entity for this in home assistant would allow you to trigger an alert so that you know when it’s offline and can go manually reset your dongle.

rmurfster:
Had the exact problem last night. This would be great.

PR #264, opened 2026-06-20
feat: Operating State sensor + Off-Grid binary sensor (#262)

### LINKED ISSUE #262: Map better the states of inverter to a sensor
### Problem or Use Case

Currently the inverter status are showing only as  "normal", but the numeric state is changing

This graph show 
- inverter 1 - this integration Status entity
- inverter 1 - this integration Status Code entity
- inverter 2 - other integration state entity

Image

As the inverters are connected in  parallel (two split-phase for 3-phase connection) is easy to see that the status is synchronized. I'm using this setup to compare both integrations each inverter on one integration.

A single entity on other integration can show the mapping of numeric status. Here there is not user friendly entity that can easily used to view the inverter status with the indications that came from status code register

Currently having "Status" and "Status Code" names gives to user a sens that it should have the same information, just numeric and string, and this is not the case now.

### Proposed Solution

Update the entity  (or create a new one if there is need to keep this as it is now) to maps the status codes as user friendly / readable information as per documentation.

Image

I don't know if this was made because on web it does not show the combined state, but having the combined state is very usefull to automations and detect when in off-grid mode.

### Alternatives Considered

creating a local template to map

### Affected Device Model(s)

All Models

btli:
## Summary

Resolves #262. The inverter's operating-mode code was only exposed as the raw numeric **Status Code** (e.g. `16`, `0`, `32`), while the text **Status** sensor showed a cloud health string ("normal") that barely changed — so there was no user-friendly, automation-ready view of what the inverter is actually doing, and the two entities were confusingly named.

This adds a friendly decode of that code (INPUT register 0 `device_status`, which is the *same value* as the cloud `status` field — so it works identically in **cloud, local, and hybrid**):

| New/changed entity | Type | Behavior |
|---|---|---|
| **Operating State** | enum sensor (primary, new) | Friendly label per EG4 "Table 9": `Battery → Grid`, `Grid → Battery`, `PV → Battery`, `Off-Grid (Battery)`, … |
| **Off-Grid** | binary sensor (new platform) | ON whenever the inverter is islanded — one boolean for off-grid automations |
| **Status → Cloud Status** | sensor (renamed) | Display-only rename to disambiguate the cloud health string; **entity_id unchanged (non-breaking)** |
| **Status Code** | sensor (unchanged) | Raw numeric kept for diagnostics / back-compat |

### Mapping (EG4 Table 9)

| Code | Operating State | Off-Grid |
|------|-----------------|:--------:|
| 0x00 Standby · 0x01 Fault · 0x02 Programming | Standby / Fault / Programming | |
| 0x04 / 0x08 / 0x0C | PV → Grid / PV → Battery / PV → Battery + Grid | |
| 0x10 / 0x14 | Battery → Grid / PV + Battery → Grid | |
| 0x20 / 0x28 | Grid → Battery / PV + Grid → Battery | |
| 0x40 | Off-Grid (Battery) | ✅ |
| 0x60 | AC-Coupled Charging | |
| 0x80 / 0x88 / 0xC0 | Off-Grid (PV, unstable) / (PV + Charging) / (PV + Battery) | ✅ |

> `0x60` is *named* "off-grid + battery charging" in Table 9 but *described* as on-grid AC-coupled charging; the description is treated as authoritative, so it is **not** flagged off-grid. Unknown codes → `unknown` (raw Status Code retained for diagnosis).

## Design

- **Shared decode** in `const/operating_state.py` (`operating_state_slug`, `is_off_grid`, `OFF_GRID_STATUS_CODES`), injected into both the LOCAL (`_build_runtime_sensor_mapping`) and CLOUD/HYBRID (`_process_inverter_object`) paths. **Integration-only — no pylxpweb change.**
- **Localizable enum**: state stored as stable slugs; names + states translated across all 12 locales. Sensor bases now skip `_attr_name` when a `translation_key` is declared, so HA resolves the localizable name from translations instead of a hardcoded English string (HA's `_attr_name` otherwise overrides `translation_key`).
- **Offline-safe** (#256 philosophy): a no-runtime inverter publishes `operating_state=None` ("unknown") so the entity is always created and never vanishes.

## Testing & validation

- `ruff` ✓, `mypy --strict` ✓, Silver/Gold (6/6)/Platinum validators ✓
- **1425 passed / 1 skipped** — 60 new tests in `tests/test_operating_state.py` (decode, both coordinator paths, offline/no-runtime path, binary sensor, name-localization, locale parity)
- Adversarial review (Codex) before ship — caught and fixed a gap where the no-runtime cloud path skipped creating the entity; now regression-tested.
- Docs: `DATA_MAPPING.md` §9 Table 9 section; `CHANGELOG.md`.

> Follow-up (out of scope): the orphaned `dongle_connectivity` binary_sensor translation (left over from the platform removed in #82) has no backing entity — a candidate cleanup.


ivanfmartinez:
> 0x60	AC-Coupled Charging

@btli I have just confirmed that 0x60 is off-grid ac-couple charging. Tuned off the main breaker when ac couple power is more than the consumed by home (battery charging).

Luxpower docs has inconsistency because one column says "off-grid" and other says "on-grid". 

docs

0x40 appear with or without ac-couple if battery is discharging in off-grid.

0x20 appear when charging and on-grid, no matter if power cames from ac-couple or not. When connected to grid and charging the only way to know that it is using ac couple is looking at ac couple power.

From the luxpower docs table I understand that all values starting with 0x40 are related to off-grid.

As a suggestion instead of using unknown when the state is unmapped why not use just the original value. Using unknown will make the need to check the numeric field, using the value will keep the real state change. And probably make easier to find newer/undocumented states.

ivanfmartinez:
> 0x20 / 0x28	Grid → Battery / PV + Grid → Battery

The key used to 0x20 is **ac_charging** and I understand that is charging from Alternate Current, but all descriptions are meading **Grid -> Battery**, my tests confirmed that this state is defined when charging also from ac-couple without power from grid,  I think that **AC Charging** or **AC -> Battery** is a better description for this key.

This appear to be another case when one column from doc is correct but other description column is incorrect.

The other integration has used that **Grid Charging Battery** and I never liked it because its not what is happening in  my system.


https://github.com/joyfulhouse/eg4_web_monitor/pull/264/changes#diff-dbccc2b9653d0a07473940464985096a963bbed057a598d126c8e9478c79c6dfR492

btli:
Thanks @ivanfmartinez — the hands-on hardware testing was exactly what was needed here. All of it is now applied (pushed in `46eff62`):

- **`0x60` is off-grid.** Your main-breaker test settles the contradiction in the Luxpower table (the *name* column says off-grid, the *description* column says on-grid). I cross-checked the integration you've been comparing against — `lxp_modbus` (`ant0nkr/luxpower-ha-integration`), which is the source of the "Battery Discharging to Load – Surplus Exporting to Grid" wording — and it also classifies `0x60` as `Off-Grid: AC Coupled PV Charging Battery`. So `0x60` → **Off-Grid (AC-Coupled Charging)**.
- **Off-grid is now a rule, not a list:** any code **≥ 0x40** (bit 6/7) is treated as off-grid, matching how every documented ≥0x40 code in the table reads. This also means an undocumented off-grid combination still trips the **Off-Grid** binary sensor instead of silently reading on-grid.
- **`0x20` / `0x28` relabeled `AC → Battery` / `PV + AC → Battery`** (not "Grid → Battery"). You're right that it also occurs charging from AC-coupled PV with no grid power — so this is one spot where we deliberately differ from `lxp_modbus`'s "Grid Charging Battery", which you mentioned never matched your system.
- **`0x11` → Standby** (the alias `lxp_modbus` also handles).
- On your "show the raw value instead of unknown" suggestion: because `Operating State` is an `enum` sensor (so the states can be localized into all 12 languages), it can't hold arbitrary numbers — but unmapped codes now emit a `DEBUG` log with the raw code, and the numeric **Status Code** sensor always carries it, so new/undocumented states are still easy to spot and report.

The friendly labels use a concise arrow style (e.g. `Battery → Grid`, `AC → Battery`, `Off-Grid (Battery)`); the meanings above are what changed. Mind giving it another look against your two-inverter setup once this lands?

ivanfmartinez:
About the descriptions of 0x10 and 0x14
they are now as :
* **Battery → Grid**
* **PV + Battery → Grid**

In reality they are the same case of the AC Couple Charging, the power goes to AC, them first to load, and the surplus goes to grid. Other integration uses this term **Battery Discharging to Load - Surplus Exporting to Grid** for state 0x10.

This state happens in my inverter when the consumption goes over the P_import value (hold-116), and probably the same will happen on peak shaving config.

Maybe we can use better name for the user understand correctly what is happening.

ivanfmartinez:
> Mind giving it another look against your two-inverter setup once this lands?

No problem to do this again when a release is available to test.

Issue #261, opened 2026-06-18
Some sensors in Hybrid mode becaming unknown/unavailable sometimes


ivanfmartinez:
### Integration Version

v3.4.0b13

### Home Assistant Version

2026.6.3

### Connection Mode

Hybrid (Local + Cloud)

### Device Model(s)

Other

### Describe the Bug

Sometimes when checking sensors I have seen they became without value.

Looking more in detail on a single sensor "Fault Code" I have seen that it became unknown many times.
Comparing with "Status Code" the status code does not appear to became unavailable.

Comparing with a Cloud only inverter I see that Fault Code appear only on the hybrid.

Not all sensors have same behavior, Grid Type for example does not became unavailable.

Battery Bank SOC, appear as fine on graph, but when I look on my log I can see failures indicating "unavailable" on a template sensor used only with the Battery Bank SOC. Each time this message is found on my log I can see before lines from the Cloud Coordinator custom_components.eg4_web_monitor.coordinator_http

### Steps to Reproduce

Don't know how to reproduce in another environment.
In my case I found it on hybrid connection inverter

The log lines are the ones that appeared when I found the missing value on my template sensor that uses battery soc

I just found that I'm not using debug for all eg4 paths in my log. Will enable for all and return with more information if I found.

### Home Assistant Logs

```shell
2026-06-18 02:19:11.200 DEBUG (MainThread) [pylxpweb.transports._register_data] [99999T9909] combined path: battery_count (reg 96) = 0
2026-06-18 02:19:11.200 DEBUG (MainThread) [custom_components.eg4_web_monitor.coordinator_http] Inverter 99999T9994 (LXP-LB-US 10K): has_data=True, _runtime=present, _energy=present, _battery_bank=present, battery_count=0, batteries_len=0
2026-06-18 02:19:11.200 DEBUG (MainThread) [custom_components.eg4_web_monitor.coordinator_http] Inverter 99999T9909 (LXP-LB-US 10K): has_data=True, _runtime=present, _energy=present, _battery_bank=present, battery_count=8, batteries_len=8
[...]

Issue #226, opened 2026-05-04
Integration fails on polling timeout


donbeal80:
### Integration Version

v3.2.0

### Home Assistant Version

2026.4.4

### Connection Mode

Hybrid (Local + Cloud)

### Device Model(s)

6000XP

### Describe the Bug

I have a bit of unique setup. Due to a lack of compute at my home, and an aversion to stringing up RPIs everywhere for one off tasks, I offload HA and the integration to an offsite server. I'm configured for hybrid and the inverter is polled via a VPN tunnel (handled by site-to-site routers, not the HA VM); I'm also configured to poll the inverter every 5sec.

The problem: I have to reload the integration in HA every time the connection is lost between inverter and HA server. Polls replicate the previous period's value until I reload the integration, see attached. Also, inverter/dongle and server IPs never change.

Is there any timeout or retry that can be introduced to fix this?

BTW, this is really a great integration, thank you! Immense value for eg4 users especially with deprecation of lxp-bridge.

### Steps to Reproduce

1. Break network connectivity between the inverter and HA
2. Note polled values get stuck rather than rolling back to zero/null
3. Reconnect network
4. Note polled values remain stuck
5. Reload HA integration
6. Polling works again

### Home Assistant Logs

```shell

```

### Screenshots

<img width="2290" height="498" alt="Image" src="https://github.com/user-attachments/assets/c9e4c768-67c4-4df9-ab97-182f3ebaa596" />

<img width="1238" height="713" alt="Image" src="https://github.com/user-attachments/assets/3cb5849c-73ff-4813-ad97-19df486eba4c" />

donbeal80:
Noting logs.

```
Logger: pylxpweb.transports.dongle
Source: runner.py:289
First occurred: 3:00:55 AM (146 occurrences)
Last logged: 3:19:54 AM

Timeout waiting for dongle response
```

```
Logger: asyncio
Source: runner.py:289
First occurred: 3:17:19 AM (63 occurrences)
Last logged: 3:19:54 AM

socket.send() raised exception.
```

btli:
Apologies for the slow response here — this one deserved a proper investigation before answering, and your VPN-tunnel setup turned out to be a really useful stress case for work that landed in the current betas.

**Triage result: partially addressed in v3.4.0-beta.2/beta.3, one piece still open.**

What's relevant in the current betas (your report is against v3.2.0, which predates all of it):

[...]

Issue #169, opened 2026-02-24
Secondary Inverter Battery Bank show multiple sensors as Unknown / Unavailable - Primary Inverter Battery Bank No problems


trinitron79:
This is what the Secondary Inverter Battery Bank shows inside HA - This data is being pulled via Modbus + wifi for Gridboss.

<img width="298" height="1080" alt="Image" src="https://github.com/user-attachments/assets/acb249d5-a743-4eaf-a79c-026cae21ca81" />

Primary Inverter Battery Bank.

<img width="288" height="1080" alt="Image" src="https://github.com/user-attachments/assets/dd7356ec-9319-421a-b882-82e5ef082569" />

Battery's are all paralleled together, inside EG4 monitor site "Share Battery" is Enabled.

Visual of battery cables + com cable
<img width="953" height="407" alt="Image" src="https://github.com/user-attachments/assets/b32002a7-f81c-4abb-8399-6c7325da6f16" />

How batteries coms are done, with Primary inverter connecting from the Bat Com port to the CAN port of battery #1.
<img width="1059" height="178" alt="Image" src="https://github.com/user-attachments/assets/cd86ad93-946a-4b83-b35d-ab01e702126b" />

[home-assistant_eg4_web_monitor_2026-02-24T19-38-46.216Z.log](https://github.com/user-attachments/files/25529374/home-assistant_eg4_web_monitor_2026-02-24T19-38-46.216Z.log)

trinitron79:
Sounds good!  I'll give Signature Solar a call JUST to confirm I'm not doing something odd, I did talk to them about this and they said there is only one connection, but worth the few minutes to confirm if I should only have a single connection or if two are fine etc.  I'll give you an update after I talk to them.

PR #265, opened 2026-06-20
fix: HYBRID battery-bank + fault-code sensors no longer flicker unknown/unavailable (#261)

btli:
## Bug (#261)

In **HYBRID** mode some sensors intermittently went `unknown`/`unavailable` while cloud-backed ones (`Status Code`, `Grid Type`) never did. Reported by @ivanfmartinez (LXP-LB-US 10K, parallel system) with a debug log showing `combined path: battery_count (reg 96) = 0` immediately before a `Battery Bank SOC` drop.

## Root cause

HYBRID rebuilds the sensor dict every poll and writes a key **only when its source value is non-None** (`_map_device_properties` / the transport overlays). So a transient local-transport gap drops the key. The two entity base classes then diverge — which is why the symptoms differ:

- A **missing key → `unknown`** for a normal `EG4BaseSensor` (its `available` ignores key presence).
- A **missing key → `unavailable`** for `EG4BatteryBankEntity` (its `available` is `key in sensors`).

Two independent triggers, both in `_process_inverter_object`:

1. **`Battery Bank SOC` → unavailable.** The bank was built from `transport_battery` but gated on `transport_battery.battery_count > 0` — i.e. **Modbus reg 96**, which is unreliable on parallel/multi-battery systems (#258/#170) and intermittently reads 0 even for a real bank (the aggregate `soc` stays valid). When it read 0 the code dropped the **whole** bank with **no cloud fallback**.
2. **`Fault Code` / `Warning Code` → unknown.** These are transport-exclusive (the cloud `getInverterRuntime` has no fault field), injected only when `transport_runtime` is present. A local Modbus link-down clears `_transport_runtime`, so they vanished.

## Fix (integration-only — no library change)

1. **Battery bank cloud fallback.** New `_bank_battery_count()` helper; the block is now `transport count > 0 → transport; elif cloud count > 0 → cloud; else skip`. A genuine shared-battery secondary reports 0 on **both** sources and stays correctly skipped (**#169 preserved**). The cloud bank is an independent source, so it's unaffected by the flaky local read (reg 96 lives in the `bms_data` group that fails silently).
2. **Sticky fault/warning codes.** When `transport_runtime is None` **and a transport is attached** (HYBRID link-down), `fault_code`/`warning_code` carry their last-known value forward from the previous poll. Scoped to status codes only — measurements stay honestly absent during an outage (**#226**). The transport gate keeps pure-CLOUD a deliberate no-op.

## Test plan

- New `tests/test_issue_261_hybrid_sensor_flicker.py` — **7 regression tests**: reg-96=0 cloud fallback, transport-count precedence, #169 secondary-skip, link-down carry-forward, carry-forward chaining/first-poll, live-overlay-wins, and the pure-CLOUD gate no-op. The two root-cause tests fail without the fix.
- Full suite: **1372 passed, 1 skipped** (pre-existing, unrelated).
- `ruff check` / `ruff format` clean; `mypy --config-file tests/mypy.ini` clean.

## Adversarial review (tri-vendor)

Claude + Codex (GPT-5.5 @ xhigh) + Gemini 3.1 Pro, two rounds. Round 1: Claude & Gemini clean; Codex raised one **MEDIUM** — the carry-forward can republish a stale code indefinitely during a *sustained* link-down. Resolved by gating to an attached transport; the **bounded-expiry** suggestion is **risk-accepted** (deliberate sticky behavior: fault codes are rare, there's no source of truth for fault state during a local outage, other sensors already signal link state, and an expiry would just reintroduce the flicker this fixes). Round 2: **all three clean.**

## Note on commits

This branch is stacked on the trivial `#258` changelog/manifest commit (`69ccd0d`); it'll drop out of this diff once #258's own PR merges into `integration/3.4.0`.


btli:
### Code review

No issues found. Checked for bugs, CLAUDE.md compliance, git history, code-comment guidance, and prior-PR review context.

Two minor documentation inaccuracies surfaced during review were corrected in 3295467 (comment said "(HYBRID/LOCAL)" though this path serves HYBRID + pure CLOUD; helper docstring over-claimed "non-negative int"). No functional issues. The #169 shared-battery-secondary skip is preserved, and the fault/warning carry-forward is correctly scoped to status codes (measurements stay absent during an outage per #226).

Issue #259, opened 2026-06-16
Actions for 12000XP


brendonlobo123:
### Problem or Use Case

besides data refresh I do not see actions like set mode, start quick charge etc.

### Proposed Solution

add actions like other inverters

### Affected Device Model(s)

12000XP

ivanfmartinez:
on my LXP inverter I have
- quick charge is a switch on the inverter
- normal and standby mdes appear as select
- some binary switches to enable/disable some other modes (ac charge, peak shaving, off grid) 
   - no complete configuration like on web interface, but the basics are available for my inverter

brendonlobo123:
Sorry am I not looking in the right place?




Book with me <https://calendar.app.google/sqymqbycKJdV8y1j6>

brendonlobo123:
<img width="1080" height="2424" alt="Image" src="https://github.com/user-attachments/assets/bf51345f-00be-4fcd-9915-997cf5237d69" />
<img width="1080" height="2424" alt="Image" src="https://github.com/user-attachments/assets/a112bd23-d585-480b-88a9-4d68339cf1a5" />

ivanfmartinez:
I have 4 blocks 
- **Controls**
- Sensors
- **Configuration**
- Diagnostics

your is missing the two blocks that have writable entities, probably a wrong/missing mapping in some place on the code, or some kind of read-only mode. I dont understand the code enough yet to help.

Your screenshot confirms also that you also have two has_data sensors as I reported on #253 

@brendonlobo123 have you tried to contact support to downgrade your dongle firmware to allow the local mode connection to see if this changes ? https://github.com/joyfulhouse/eg4_web_monitor/issues/250#issuecomment-4698243404

brendonlobo123:
I asked.



Book with me <https://calendar.app.google/sqymqbycKJdV8y1j6>

brendonlobo123:
@btli is this an easy fix?

btli:
Thanks for the detailed report and screenshots — they pinned this down exactly. 🙏

## What was happening

@ivanfmartinez was right that it was a mapping problem. Your 12000XP reports its model to the EG4 cloud as **`SNA-US 15K`** (visible in your second screenshot under *Device info*). The integration was deciding which inverters get writable entities (switches, numbers, selects) by matching that model name against a fixed list of known names — and `SNA-US 15K` matched **none** of them, so your inverter was created with **no Controls and no Configuration sections at all**.

[...]

PR #283, opened 2026-07-02
feat: expose AC Charge schedule as time entities (#277)

### LINKED ISSUE #277: Configure AC Charge Start/Stop Times
### Problem or Use Case

It appears that the AC Charge start and end times are not exposed in home assistant. I have dual rate power with the rate being about $0.21 per kwh from 8:00 am to 8:00 pm weekdays. Off peak is around $0.10. I have my 12000XP configured to ac charge from 8a to 8p so that it runs battery overnight. My excess solar in the summer covers the charge time. In winter when I have no banked solar power, I will reverse the process with batteries used during the day and charging at night. Since there is about 5% power loss going through the charge / discharge process, I will turn off battery use on the weekends. Currently I need to manually go in to the maintenance and change the times to 24 hours (00:00 to 23:59) on Friday and reset it back on Sunday night. 

### Proposed Solution

If AC Charge times were exposed, the logic for all the battery usage could be automated. I will be investigating programming this myself on a separate server but it would be nice if it were integrated. If you want me to attempt to code this as a separate pull on the project, let me know.

### Alternatives Considered

I was also wondering if I could set discharge currents to 1 through the automation? That sends an EPS Over load notice and does seem to stop the battery use. I am not sure how long it takes to reset it again. That approach would require the ac charge times to be manually changed in spring and fall however. Since my 12000XP is not showing any configuration values yet (is the 15k model), I cannot experiment yet.

### Affected Device Model(s)

All Models

btli:
## Summary

Adds a native Home Assistant **`time` platform** exposing the AC Charge schedule (issue #277): **AC Charge Start/End Time** for all **three daily windows**, per control-capable inverter. The reporter (12000XP, dual-rate power) can now automate his weekday peak/off-peak charging window and the weekend 00:00-23:59 flip instead of editing times by hand in the EG4 portal.

- **Window 1** entities enabled by default; **windows 2 and 3** created with `entity_registry_enabled_default=False` (most users schedule one window — enable the others from the entity registry).
- Values follow **parameter polling in every mode** (CLOUD/LOCAL/HYBRID), so changes made in the portal appear in HA.
- The existing **AC Charge Enable** switch (HOLD 21 bit 7) and AC charge type (reg 120) are untouched — these entities manage only the schedule times.

## Register evidence (do-not-guess proof)

**Registers 68-73 = three windows × (start, end), one PACKED time per register.**

| Reg | Boundary | Cloud params returned when reading *this one register* |
|-----|----------|--------------------------------------------------------|
| 68 | Window 1 start | `HOLD_AC_CHARGE_START_HOUR` + `HOLD_AC_CHARGE_START_MINUTE` |
| 69 | Window 1 end | `HOLD_AC_CHARGE_END_HOUR` + `HOLD_AC_CHARGE_END_MINUTE` |
| 70 | Window 2 start | `HOLD_AC_CHARGE_START_HOUR_1` + `HOLD_AC_CHARGE_START_MINUTE_1` |
| 71 | Window 2 end | `HOLD_AC_CHARGE_END_HOUR_1` + `HOLD_AC_CHARGE_END_MINUTE_1` |
| 72 | Window 3 start | `HOLD_AC_CHARGE_START_HOUR_2` + `HOLD_AC_CHARGE_START_MINUTE_2` |
| 73 | Window 3 end | `HOLD_AC_CHARGE_END_HOUR_2` + `HOLD_AC_CHARGE_END_MINUTE_2` |

Source: the **live cloud register probe** on a FlexBOSS21 (pylxpweb `docs/inverters/FlexBOSS21_52XXXXXX78.json`, `startRegister=N, pointNumber=1`). Getting **two** named params back from **one** 16-bit register is only possible if the register packs both fields — and it also pins the cloud naming: window 1 unsuffixed, windows 2/3 suffixed `_1`/`_2`. This matches pylxpweb's `SCHEDULE_CONFIGS` (`ScheduleType.AC_CHARGE` → base register 68) and its cloud schedule helpers (`endpoints/control.py`), which the EG4 portal's own writes use.

### Byte-packing proof (hour-low / minute-high)

pylxpweb `constants/scaling.py` (the codepath its own Modbus schedule writes use):

```python
return (hour & 0xFF) | ((minute & 0xFF) << 8)   # pack_time — "hour in low byte, minute in high byte"
>>> pack_time(23, 30)
7703                                             # 0x1E17: 0x1E=30 (min, high), 0x17=23 (hour, low)
```

The integration imports `pack_time`/`unpack_time` from pylxpweb (single source of truth) rather than re-implementing the byte order; round-trip is pinned by tests (00:00→0, 23:59→15127, 23:30→7703, 08:00→8, 00:30→7680).

### Known pylxpweb naming wart (documented, handled)

pylxpweb's `REGISTER_TO_PARAM_KEYS` still carries a stale pre-probe interpretation of 68-73 (one field per register: `HOLD_AC_CHARGE_START_HOUR_1` @ 68 … `HOLD_AC_CHARGE_ENABLE_1/2` @ 72/73). Local `read_named_parameters()` therefore surfaces the **raw packed values under those misleading names**. The integration maps each register to an alias chain (`LOCAL_AC_CHARGE_TIME_PARAM_KEYS`, with fallbacks for a future pylxpweb rename/removal) and unpacks the value itself, with hour≤23/minute≤59 sanity validation (corrupt reads → entity unavailable, never a bogus time). Correcting the pylxpweb table is deliberately **not** done here — `HybridInverter.get_ac_charge_status()` and any external consumer read those names today; that is follow-up library work.

## Write paths

- **LOCAL / HYBRID with attached transport**: one packed register write per changed boundary via the new `coordinator.write_register()` (raw-register sibling of `write_named_parameter()`, with the same reconnect handling). Single-register writes go out as **FC06** — pylxpweb's own schedule helper documents that the firmware rejects FC16 multi-writes on these registers.
- **CLOUD**: the portal's own named-parameter writes — `HOLD_AC_CHARGE_{START|END}_HOUR{suffix}` + `..._MINUTE{suffix}` via `control.write_parameter()` (exactly what pylxpweb's verified cloud `set_ac_charge_schedule()` sends, scoped to the one changed boundary).
- Optimistic UI value during the write; parameters re-read afterwards so all six entities converge.
- Overnight windows (end < start, e.g. 20:00 → 08:00) are firmware-legal and **not** cross-validated.

## Family gating (12000XP included)

Entity creation uses `is_supported_control_model()` — the #259 family-aware gate (model-substring OR detected family in {EG4_OFFGRID, EG4_HYBRID, LXP}) — so the reporter's unit (cloud model "SNA-US 15K", EG4_OFFGRID family) gets the entities in every mode. The LOCAL/HYBRID static first-refresh already populates `features.inverter_family` via `_features_from_family()`, which this gate consumes, so entities are created on the zero-read static phase too. Positive tests cover SNA/EG4_OFFGRID, EG4_HYBRID, LXP; negative tests cover unknown models and GridBOSS.

## Changes

- `custom_components/eg4_web_monitor/time.py` — new platform (structure mirrors number.py)
- `base_entity.py` — `EG4BaseTime` + `optimistic_time_context` (time-typed mirror of `EG4BaseNumber`)
- `coordinator.py` — `write_register()` helper
- `const/modbus.py` — `AC_CHARGE_SCHEDULE_BASE_REGISTER`, `LOCAL_AC_CHARGE_TIME_PARAM_KEYS` (+ exports)
- `__init__.py` — `Platform.TIME` registered (setup + unload via PLATFORMS)
- `strings.json` + **all 13 locales** — `entity.time` names (translation_key only, no `_attr_name`, per #262)
- `docs/DATA_MAPPING.md` — new "AC Charge Time Schedule Registers" section with the evidence and the local-cache naming caveat
- `CHANGELOG.md` — [Unreleased] entry
- `tests/test_time_entities.py` — 54 tests (TDD: written first)

A small preparatory commit fixes three pre-existing strict-mypy errors surfaced by HA 2026.1 stubs (present on the base branch with a fresh latest-HA env; they block the pre-commit mypy hook for any change). No suppressions — typed locals + removal of a now-unused ignore.

## Testing

- `pytest tests/` — **1574 passed, 1 skipped** (54 new)
- `mypy --config-file tests/mypy.ini custom_components/eg4_web_monitor/` — **0 errors** (strict)
- `ruff check` + `ruff format` — clean
- `tests/validate_translations.py` — all 13 locales complete (658 keys)
- Silver/Gold/Platinum tier validators — pass

New-test coverage: packing round-trip incl. byte-order pins; entity creation per family (positive: FlexBOSS21, SNA-US 15K/EG4_OFFGRID, EG4_HYBRID, LXP; negative: unknown model, GridBOSS); windows 2/3 disabled-by-default; local packed write (register + packed value asserted); cloud named writes (params + values asserted); cloud write failure → `HomeAssistantError`; no-transport-no-client → error; midnight-crossing end<start accepted; optimistic value during write; seconds dropped; unavailable-until-params-polled; garbage packed value and out-of-range/non-numeric cloud values → unavailable; `Platform.TIME` in PLATFORMS.

## Verification asks for @mjstrand (and other testers)

1. Install this branch (HACS custom repository → this branch, or copy `custom_components/eg4_web_monitor`), restart HA.
2. On your 12000XP device page you should see **AC Charge Start Time 1** and **AC Charge End Time 1** under *Configuration* (windows 2/3 are available but disabled by default). Please confirm they show the same times as the portal's Maintenance page (8:00 AM / 8:00 PM in your setup).
3. Change a time in HA → confirm the portal shows the new value (and vice-versa after the next parameter poll, up to ~1 h in local modes or a few minutes cloud).
4. Try your weekend automation: set start 00:00 / end 23:59 Friday night, restore Sunday night — confirm the inverter honors the window (battery does not discharge during the charge window).
5. If you run LOCAL or HYBRID (Modbus/dongle), please also confirm a write while charging is active behaves (the write is a single register — no full schedule rewrite).

## Residual risks

- The LOCAL read depends on pylxpweb's current (stale) names for registers 68-73; a future pylxpweb rename is covered by the alias chains + plain-address fallback, and values are sanity-validated. Tests pin today's behavior.
- Cloud hour/minute are two sequential writes (portal semantics); a failure between them leaves the boundary mixed until the retry — same exposure as the portal itself.
- Register 68-73 packing is verified on EG4_HYBRID hardware (FlexBOSS21 probe) and is the LXP-protocol-wide schedule layout; 12000XP (SNA platform) confirmation is exactly what the reporter verification asks cover.

Fixes #277


btli:
Addressed the adversarial review verdict (FIX-THEN-MERGE) — head is now `43a9290`, rebased onto `integration/3.4.0` @ `1c2df85` (post-#286; strings.json/locales/CHANGELOG merged cleanly with the block-size additions).

**P1 (cloud partial write)** — any failure after the first successful cloud write now runs a best-effort parameter refresh *before* re-raising, and the error states the schedule time may be partially applied and that device state was re-read. The entity converges to the actual (mixed) device value instead of showing the stale pre-write time. Covered for both `success=False` and raised-exception failures; a failed *first* write does not refresh (nothing changed on the device).

**P2 (write ok, refresh failed)** — the optimistic value is now retained when the write landed but the follow-up refresh failed, and clears once fresh parameter data arrives (cache decodes to the written value or any value other than the pre-write one). Coordinator ticks still carrying the stale pre-write value do not clear it. The finally-always-clears `optimistic_time_context` helper is removed in favor of explicit management.

Also includes `32ad071` (recorder `get_instance` import from `homeassistant.helpers.recorder` — type-checks on both HA 2025.11 and 2026.1).

Gates on `43a9290`: **1606 passed, 1 skipped** (+8 new P1/P2 tests, 62 time-platform tests total) · mypy strict **0 errors** (40 files) · ruff check/format clean · translations validator 13/13.

Issue #252, opened 2026-06-13
Migrating to Cloud to Hybrid duplicated battery entities


ivanfmartinez:
### Integration Version

v3.4.0b10

### Home Assistant Version

2026.6.3

### Connection Mode

Hybrid (Local + Cloud)

### Device Model(s)

Other

### Describe the Bug

Previously I was using cloud only, the battery devices are created with battery serial numbers.

Today I have migrated this inverter to hybrid mode, and not it have created new battery devices numbered as 01..08 and left old devices without updating.

Probably keeping the batteries with serial number is better to best identify them, but if the new behaviour is use sequential numbers it should remove and if possible migrate the old entities.

### Steps to Reproduce

1. Configure device as cloud only
2. Change device to connect to local inverter (I used dongle)

### Home Assistant Logs

```shell
Did not saved the log when the migration was made.
```

### Screenshots

<img width="527" height="425" alt="Image" src="https://github.com/user-attachments/assets/24a90c45-17d7-4bcb-8799-21819c164ee7" />

btli:
**v3.4.0-beta.18 is out** with the battery-identity unification. On upgrade you should see, in the logs, a one-time registry migration: your positional battery devices re-identified in place to serial-first keys (automations/dashboards/history preserved), and the cloud→hybrid duplicates removed.

Verification asks for your LXP setup:
1. After upgrade + one restart: exactly **one** device per physical battery (no `-01…-NN` duplicates alongside serial-keyed ones)?
2. Switch cloud↔hybrid once more: still no new duplicates?
3. Spot-check one battery's history survived the migration.

One caveat that matters for your 8-battery rotating bank: packs that rotate more batteries than the 4 register slots are deliberately excluded from the automatic positional-history rename (attribution would be guesswork) — those old positional entities become removable orphans instead. Report back and we'll close.

ivanfmartinez:
@btli b18 took long time on startup and got error about pylxpweb version

```
2026-07-02 13:41:12.254 WARNING (SyncWorker_0) [homeassistant.loader] We found a custom integration eg4_web_monitor which has not been tested by Home Assistant. This component might cause stability problems, be sure to disable it if you experience issues with Home Assistant
2026-07-02 13:45:35.844 ERROR (MainThread) [homeassistant.setup] Setup failed for custom integration 'eg4_web_monitor': Requirements for eg4_web_monitor not found: ['pylxpweb>=0.9.36b20'].

```

[...]

PR #293, opened 2026-07-02
fix: carry once-published batteries forward across transient merge gaps (#258)

btli:
## Root cause (beta.18 DEBUG log, @ivanfmartinez's 2×LXP / 8-battery HYBRID rig)

**The local side never lost a battery.** pylxpweb's serial-keyed accumulator reports `Battery accumulator: 8 batteries populated (slots [0..7])` on **every** read through both drop windows — 818 consecutive times after the second firmware page first appeared at 14:20:44 (log lines 4895 and 10893 bracket the two drops). The raw pages show the firmware alternating two 4-battery pages and pinning to all-empty pages for long stretches (`RR raw page: 0=empty 1=empty 2=empty 3=empty (reg96=8)`, e.g. line 10892) — exactly the population the accumulator exists for, and it did its job.

**The loss is in the integration's HYBRID merge.** Battery entity availability is literally key-presence in `device_data["batteries"]` (`base_entity.py` `EG4BatteryEntity.available`), and the HYBRID branch of `_process_station_data()` rebuilds that dict **from the cloud payload as the baseline** every cycle — the transport data is only an *overlay on cloud-present batteries* (`bat_serial not in cloud_by_serial → continue`). A battery the fresh cloud `getBatteryInfo` momentarily omits (or re-keys) vanishes from the merged dict even while the local accumulator holds fresh-enough data for it.

Both reporter drop timestamps land seconds after a **fresh** (non-cached) `getBatteryInfo` POST:

| drop (reporter) | fresh cloud POST | log line |
|---|---|---|
| 16:06:45 | 16:06:44.227 | 4905 |
| 17:58:27 | 17:58:23.124 | 10888 |

The rig's rotating >4-slot bank feeds the **cloud** through the same firmware page rotation the local reads see, so a cloud payload that momentarily carries a page-grouped subset is exactly what produces "batteries 01–04 unavailable at one time, 05+ at another".

### The "duplicate serials in the accumulator" observation is a logging artifact

The 17:58:26 accumulator dump (lines 10893–10902) shows `Pos 4`/`Pos 6` both as `…BCW?017` and `Pos 3`/`Pos 7` both as `…BCW?016` — but the `Pos N:` debug decodes only register offsets 17–23 (14 chars), while the accumulator identity and `BatteryData.serial_number` use `read_battery_serial` (offsets 17–24, 15 chars: the final serial character lives in offset 24's **low byte** — visible as `offset24=0x0437` → `'7'`). The raw-page lines print the full identities: `0177 / 0172` and `0165 / 0167` — **8 distinct serials, no collapse**. The companion pylxpweb PR fixes the misleading dump and adds a real within-page duplicate-serial guard as defense-in-depth.

## Fix

**`_apply_battery_carry_forward()`** (`coordinator_http.py`), called from every battery branch of `_process_station_data()` (HYBRID merge, LOCAL-only round-robin, CLOUD-only, and the no-bank / no-batteries early exits):

- A battery **once published** for an inverter keeps its last-known mapping when a cycle's merge output omits it — availability never flaps. Carried mappings keep their original `battery_last_seen` / `battery_last_polled` stamps, so staleness stays visible **as data**, never as availability (the #261/#282 sticky precedent).
- **Identity guards** (no #252 regressions):
  - legacy positional aliases of batteries published this cycle are excluded (carrying one would re-mint a positional entity and permanently block the #252 registry migration via the "legacy key still active" gate);
  - a cached key whose `battery_serial_number` is already published under a different key is superseded, not doubled (cloud re-keying a pack must not show it twice).
- **LOCAL path** (`coordinator_local.py`): the round-robin cache is now also served when `transport_battery` is `None` or reg 96 transiently reads 0 (previously only when the batteries *list* was empty). A genuine shared-battery secondary never populates the cache, so nothing is fabricated for it.

## Tests

- `TestBatteryCarryForward` (4 tests): cloud payload shrink keeps all batteries with original `battery_last_seen`; cloud bank + transport both gone keeps all; re-keyed identity is superseded under both guard paths.
- `TestBatteryRRCacheFallback` (+2): reg 96 = 0 and `transport_battery=None` with a populated rr cache.
- Full suite: **1653 passed** (baseline 1650), ruff + mypy strict clean. The single `test_fast_with_released_lib_stays_conservative` failure is a pre-existing environment artifact (that test hard-asserts the installed pylxpweb is ≤ b19 without block-size support; this env installs b20 per the beta.18 requirement) — it fails identically on a clean `integration/3.4.0` checkout.

## Honest residual uncertainty

The reporter's log has DEBUG only for `pylxpweb.*`, not `custom_components.eg4_web_monitor`, and cloud response bodies aren't logged — so the exact shape of the shrunken/re-keyed cloud payload at the two drops can't be read directly from the log; it is inferred from the airtight elimination of the local path plus the drop↔fresh-POST correlation (2 drops in ~4h, both within ~4 s of one of the ~130 fresh POSTs). The carry-forward is deliberately mechanism-agnostic: **any** transient merge-set shrinkage (cloud omission, re-key, bank fetch failure, link-down) is now covered.

## Verification asks for @ivanfmartinez

1. Run with this build and confirm no battery entity flips unavailable across ≥24 h (the drops previously appeared within ~2–4 h windows).
2. If possible, also enable `custom_components.eg4_web_monitor: debug` — a `Carrying forward N batteries missing from this cycle` line at a former drop time confirms the mechanism directly.
3. The `battery_last_seen` attribute on each battery shows honest data age — rotation staleness will surface there instead of as unavailable entities.

Companion pylxpweb PR: duplicate-serial disambiguation + full-serial debug dump (defense-in-depth; no version bump).

Fixes remaining #258 battery-entity gaps.

---

## Review-round fixes (Opus MERGE+MERGE, Codex FIX-THEN-MERGE×2 — all findings addressed)

**Bounded eviction (both reviewers — the immortal-key problem).** Carried keys now age out: a cached mapping whose `battery_last_seen` exceeds **`BATTERY_CARRY_FORWARD_MAX_AGE` = 6 hours** is evicted with one INFO, so a physically removed pack converges without an HA restart instead of feeding frozen SoC into bank aggregates and automations forever. Rationale for 6 h: it is orders of magnitude above the seconds-to-minutes cloud/transport gaps carry-forward absorbs, and it is safe versus the 2026-06-28 **9-hour** firmware page pinning — during pinning the pylxpweb accumulator still *serves* the batteries as fresh current-cycle data, so cached-serve is not what holds them; only a battery gone from BOTH the cloud payload and the accumulator ages against the bound. The same bound now gates the LOCAL `_battery_rr_cache` re-serve path (Codex confirmed it re-inserted removed packs too). Test: physical-removal convergence within the bound, both HYBRID and LOCAL.

**Codex P1 — LOCAL fallback carried without exclusions.** A no-serial pack cached positionally (`INV-01`) was resurrected after the rr merge retired it for the arriving serial (its cached mapping has no serial for the supersede guard). Retirement is now authoritative across both sticky layers — `_merge_round_robin_batteries()` purges the retired key from the carry-forward cache — and the fallback branch passes the legacy-alias exclusion set. Test: the exact scenario.

**Opus P2 — supersede snapshot.** `current_serials` now grows as carried mappings are admitted, so two cached keys sharing one serial can never both be carried. Test included.

**Cross-layer guard for the pylxpweb `@posN` reconciliation** (companion PR): integration-level test that a transient duplicate-serial poll does not leave a lasting extra battery entity (collision key retired on the next clean poll, not resurrected).

**Inherited train fix (separate commit):** `test_fast_with_released_lib_stays_conservative` asserted against the installed pylxpweb version (failed under the manifest-mandated b20); it now simulates the feature-less library by patching `input_block_size_kwargs()` → `{}`.

**Informational (Opus):** for a *genuine* duplicate-serial pack pair the preservation win exists only at the pylxpweb layer — the serial IS the HA battery identity, so HA can only ever represent such twins as one device plus a positional collision key. Inherent to serial-keyed identity, documented here for the record.

Updated gates: **1659 passed** (1653 + 5 new review-round tests + the rehabilitated block-size test), 0 failed; ruff + mypy strict clean.

### DIFF
--- a/custom_components/eg4_web_monitor/coordinator.py
+++ b/custom_components/eg4_web_monitor/coordinator.py
@@ -358,6 +358,16 @@ class EG4DataUpdateCoordinator(
         # duplicate/misreported battery serial).  Sticky until restart;
         # positional registry rows are left untouched as orphans.
         self._battery_migration_suppressed: set[str] = set()
+        # Last published battery mapping per inverter (#258 beta.18): battery
+        # entity availability is key-presence in device_data["batteries"], and
+        # the HYBRID/CLOUD paths rebuild that dict from the cloud payload as
+        # the BASELINE every cycle — a battery the cloud momentarily omits
+        # (rotating >4 packs feed the cloud through the same firmware page
+        # rotation) vanished and its entities flipped unavailable.  A battery
+        # once published is carried forward with its last-known data instead;
+        # staleness stays visible via battery_last_seen, never via
+        # availability flapping.  Outer key = inverter serial.
+        self._battery_carry_forward: dict[str, dict[str, dict[str, Any]]] = {}
 
         # Shared-battery secondary inverters: in parallel systems the CAN bus
         # connects only to the primary.  Secondaries (role >= 2) with
--- a/custom_components/eg4_web_monitor/coordinator_http.py
+++ b/custom_components/eg4_web_monitor/coordinator_http.py
@@ -10,6 +10,7 @@ self._http_polling_interval, etc.) accessed via mixin protocol.
 import asyncio
 import logging
 import time as _time
+from collections.abc import Collection
 from datetime import timedelta
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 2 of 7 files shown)


ivanfmartinez:
> * Run with this build and confirm no battery entity flips unavailable across ≥24 h (the drops previously appeared within ~2–4 h windows).

Just updated and restarted, tomorrow I will check the entities

> * If possible, also enable `custom_components.eg4_web_monitor: debug` — a `Carrying forward N batteries missing from this cycle` line at a former drop time confirms the mechanism directly.

sorry, my bad, I used an wrong regex to filter, the eg4_web_monitor entries was on original log, but not in the filtered file. I have corrected the filtering command and next time it will be sent.

Issue #282, opened 2026-07-02
3.4.0b17 some entities became unknown


ivanfmartinez:
### Integration Version

v3.4.0b17

### Home Assistant Version

2026.6

### Connection Mode

Hybrid (Local + Cloud)

### Device Model(s)

Other

### Describe the Bug

Updated to v.3.4.0b17

Some entities became unknow few minutes ago, here one example

<img width="439" height="612" alt="Image" src="https://github.com/user-attachments/assets/138c4573-629f-452f-83f7-15a8df80cb2d" />

### Steps to Reproduce

1. inverter connected in hybrid mode

### Home Assistant Logs

```shell
2026-07-01 20:51:16.073 DEBUG (MainThread) [pylxpweb.client] API POST /WManage/api/inverter/getInverterRuntime (data_keys=['serialNum'])
2026-07-01 20:51:16.073 DEBUG (MainThread) [pylxpweb.client] API POST /WManage/api/battery/getBatteryInfo (data_keys=['serialNum'])
2026-07-01 20:51:16.074 DEBUG (MainThread) [pylxpweb.client] API POST /WManage/api/inverter/getInverterRuntime (data_keys=['serialNum'])
2026-07-01 20:51:17.612 DEBUG (MainThread) [pylxpweb.client] API POST /WManage/web/config/quickCharge/getStatusInfo (data_keys=['inverterSn'])
2026-07-01 20:51:17.612 DEBUG (MainThread) [pylxpweb.client] API POST /WManage/web/config/quickCharge/getStatusInfo (data_keys=['inverterSn'])
2026-07-01 20:51:17.806 DEBUG (MainThread) [pylxpweb.client] API POST /WManage/web/maintain/remoteRead/read (data_keys=['inverterSn', 'startRegister', 'pointNumber', 'autoRetry'])
2026-07-01 20:51:17.817 DEBUG (MainThread) [pylxpweb.client] API POST /WManage/web/maintain/remoteRead/read (data_keys=['inverterSn', 'startRegister', 'pointNumber', 'autoRetry'])
2026-07-01 20:51:23.726 DEBUG (MainThread) [pylxpweb.transports._register_data] [?????T??09] combined path: battery_count (reg 96) = 8
2026-07-01 20:51:23.829 DEBUG (MainThread) [pylxpweb.transports._register_data] [?????T??94] combined path: battery_count (reg 96) = 0
[...]

PR #312, opened 2026-07-03
feat: Generator, Off-Grid and Peak Shaving schedule time entities; all schedule times disabled by default

btli:
## Summary

Adds three schedule time-entity families as native HA `time` entities and flips **all** schedule time entities (the three new families plus the existing four) to registry-disabled by default.

| Family | Registers | Windows | Family gate |
|--------|-----------|---------|-------------|
| Peak Shaving | 209-212 | 2 | EG4_HYBRID |
| Generator Charge | 256-259 | 2 | EG4_HYBRID or EG4_OFFGRID |
| Off-Grid | 269-274 | 3 | EG4_HYBRID |

Requires the matching pylxpweb change: **joyfulhouse/pylxpweb#209**.

## What's new vs the existing four families

The `ScheduleTimeSpec` table (const/modbus.py) gained per-family fields mirroring pylxpweb's `ScheduleConfig`, and a drift-guard test pins the two together:

- **`windows`** (2 or 3) — replaces the hardcoded 3-window assumption.
- **`bare_first_window`** — these families number ALL windows `_1.._N` (no bare window), unlike the classic `("", "_1", "_2")`.
- **`write_via_time_api`** — cloud writes use pylxpweb's atomic `write_time_parameter` (one boundary call sets hour+minute together, no partial-write re-read). The #283 convergence machinery for the existing four families is untouched.
- **`read_lsp_base`** — Peak Shaving reads back under the interleaved `LSP_HOLD_DIS_CHG_POWER_TIME_37..44` params.

`EG4ScheduleTimeEntity` derives its window suffix, cloud read params, and composite writeTime param from the spec. The new entities are feature-detected on `ControlEndpoints.write_time_parameter` and skipped (logged once) on older pylxpweb — **the manifest floor bump lands at release, not in this PR**.

## Disabled by default (explicit product decision)

Every schedule time entity — the three new families **and** the existing four (AC Charge/First, Forced Charge/Discharge), **window 1 included** — now gets `entity_registry_enabled_default=False`. Schedules are an advanced opt-in feature. Users who already enabled/kept a window keep it (the registry persists; the default only affects new registrations).

## LOCAL polling

Two EG4_HYBRID-gated reads are added — `(209, 4)` for Peak Shaving and one combined `(256, 19)` for Generator + Off-Grid — family-gated per-cycle exactly like the existing `(152, 6)` AC First read, so non-hybrid firmware that NAKs them can't loop the #282 early-retry. (Generator's *entity* gate also allows EG4_OFFGRID, but the LOCAL read is hybrid-only per the register scope verified so far; offgrid Generator relies on cloud/hybrid.)

## SNA Generator evidence

The SNA12K-US register probe (`pylxpweb docs/inverters/SNA12KUS_52XXXXXX68.json`, blocks 255-259) reports `HOLD_GEN_{START|END}_{HOUR|MINUTE}_{1,2}` at regs 256-259 → **positive**, so Generator's entity gate includes EG4_OFFGRID. Off-Grid and Peak Shaving time params are **absent** on that dump → EG4_HYBRID only.

## i18n

`strings.json` + all 13 translations get names for the 14 new entity keys, following the existing schedule-time naming convention per language.

## Tests

`test_time_entities.py` (200) + `test_coordinator_local.py` (107) updated/extended: spec-table drift guard vs pylxpweb `SCHEDULE_CONFIGS`, suffix + register math for all three bases, disabled-by-default for every schedule entity, family gating (hybrid; Generator also offgrid; LXP/unknown fail closed), the writeTime cloud write path, the Peak Shaving LSP read, LOCAL range gating per family, and older-pylxpweb feature detection. Full suite: **1849 passed, 3 skipped**; ruff + mypy strict clean.

## Review fix round (cfcc257)

- writeTime cloud writes now wrap `write_time_parameter` in `try/except → HomeAssistantError` (Codex-confirmed: a persistent DATAFRAME_TIMEOUT surviving pylxpweb's retries would otherwise escape `async_set_value` raw).
- Combined `(256, 19)` LOCAL read split into `(256, 4)` + `(269, 6)` to skip the deliberately-unmapped `260-268` zone.
- Generator `(256, 4)` added to the EG4_OFFGRID read branch (SNA probe confirms those regs), closing the Generator readback gap on LOCAL-only SNA. Net LOCAL schedule reads: EG4_HYBRID `(209,4)+(256,4)+(269,6)`; EG4_OFFGRID `(152,6)+(256,4)`; grid-tied none.

**1851 tests pass** (was 1849); ruff + mypy strict clean.

## Should-verify at ship time (accepted from review)

- **Live cloud-mode smoke for the Peak Shaving LSP page boundaries** — confirm the portal returns `LSP_HOLD_DIS_CHG_POWER_TIME_37..44` for both windows and that a write→readback round-trips through the entity.
- **Off-Grid-schedule-on-grid-tied-HYBRID portal check** — confirm the FlexBOSS21 portal actually exposes the Off-Grid schedule section (registers verified; portal UI surface unverified) before relying on the cloud read/write for that family.


btli:
## Review fix round (cfcc257)

Applied the required fix + two range items from the Codex fix-then-merge / Opus review:

1. **REQUIRED — writeTime error handling**: `_async_write_cloud`'s writeTime branch now wraps `write_time_parameter` in `try/except → HomeAssistantError` (same as the classic branch), so a failure surviving pylxpweb's retries (persistent DATAFRAME_TIMEOUT) surfaces as `HomeAssistantError` instead of a raw `LuxpowerAPIError`/`LuxpowerConnectionError` out of `async_set_value`. No partial-write re-read — the atomic call has no mixed-time window. Added a test that RAISES (the prior test only covered `success=False`).
2. **Range hygiene**: the combined `(256, 19)` LOCAL read is split into `(256, 4)` + `(269, 6)` to skip the deliberately-unmapped `260-268` zone (model-dependent `HOLD_EXPORT_LOCK_POWER` etc., pylxpweb `inverter_holding.py`) — 9 wasted reads/poll avoided.
3. **Generator offgrid readback**: added `(256, 4)` to the EG4_OFFGRID branch (the SNA probe proves regs 256-259 exist), closing the write-works-but-readback-missing gap for Generator entities on LOCAL-only SNA. Net LOCAL schedule reads: **EG4_HYBRID** = `(209,4)+(256,4)+(269,6)`; **EG4_OFFGRID** = `(152,6)+(256,4)`; grid-tied = none.

**1851 tests pass**; ruff + mypy strict clean.

### Accepted should-verifies (at ship time)
- **Live cloud-mode smoke for the Peak Shaving LSP page boundaries** — confirm the portal returns `LSP_HOLD_DIS_CHG_POWER_TIME_37..44` for the two windows and that write→readback round-trips through the entity.
- **Off-Grid-schedule-on-grid-tied-HYBRID portal check** — confirm the FlexBOSS21 portal actually exposes the Off-Grid schedule section (registers verified; portal UI surface unverified) before relying on the cloud read/write for that family.

Issue #472, opened 2026-07-17
AC Couple local register path: pin reg 179 bit 11 (FUNC_AC_COUPLING_FUNCTION)


btli:
## Summary
Give the AC Couple controls (#471 switch, #352 SOC window follow-on) a LOCAL Modbus path by pinning `FUNC_AC_COUPLING_FUNCTION` to holding register 179 bit 11.

## Evidence so far (ivanfmartinez, #471)
- The Luxpower Modbus doc places the function at **HOLD reg 179 bit 11**.
- Another integration uses that mapping, and the reporter has driven the control through it before on his LXP.
- Live named-read correlation: his LXPUS810K dump (AC couple **enabled** full-time) reads the cloud param `True`; the SNA12K-US probe unit (disabled) reads `False`.

This clears the "candidate" bar but not this project's **pin standard**: a raw↔named lockstep toggle (the procedure that pinned reg 179 bit 3 / `FUNC_PV_SELL_TO_GRID_EN` on 2026-06-12 — flip the named param, verify the raw reg-179 XOR is exactly the one bit, flip back, re-read).

## Work
1. **Pin**: run the lockstep toggle on live hardware — ivanfmartinez has offered to test on his LXP once the #471 switch is released (a raw reg-179 read before/after toggling the new HA switch is sufficient); alternatively an authorized live-write window on the dev FlexBOSS21.
2. **pylxpweb**: add bit 11 to the reg-179 decode + named local write (RMW), mirroring the bit-3/bit-7 entries.
3. **eg4_web_monitor**: let the AC Couple switch use the local path (LOCAL-only setups gain the entity; HYBRID gains local-first state/writes). The dedicated `ac_couple_soc` store stays for the SOC window (still no register for those params).

Blocked on step 1 hardware evidence. Refs #471, #352; see the reg-179 comment block in pylxpweb `constants/registers.py`.

ivanfmartinez:
@btli tested beta 2.


- The control works fine to enable/disable the ac couple
- remember that I have TWO split-phase inverters that work together to form a 3-phase system
- the expected behavior on inverters happened (both turn off or both turned on changing on single inverter), I did not remembered of that until testing today (was a long time that I have tested)
- the delay is big to get the data updated on other inverter 
- I did not forced refresh
[...]

Issue #471, opened 2026-07-17
AC Couple enable/disable switch (FUNC_AC_COUPLING_FUNCTION)


btli:
## Summary
Add an **AC Couple** switch entity per inverter that enables/disables the AC-coupled source on the smart port via the portal's `FUNC_AC_COUPLING_FUNCTION` function param.

Follow-up from #352: @mjstrand suggested toggling the function outright is a better fit for the grid↔smart-port transfer script than driving the Start/End SOC window ("it will drop the ac couple power at any battery level"), and @ivanfmartinez independently confirmed the portal uses `FUNC_AC_COUPLING_FUNCTION` (true/false) on his on-grid hybrid LXP alongside the same `_12K_HOLD_AC_COUPLE_{START,END}_SOC` params.

## Design (mirrors #352's cloud-only architecture)

**pylxpweb** (target 0.9.39b3):
- Extend `get_inverter_ac_couple_soc_limits` to also return `"enabled": bool | None` — `read_device_parameters_ranges` already returns `FUNC_*` params as booleans, so this is one more parse from the read the eg4 coordinator already makes (no extra cloud call). Absent param → `None`.
- Named setter `set_inverter_ac_couple_enabled(inverter_sn, enabled)` wrapping the existing generic `control_function(sn, "FUNC_AC_COUPLING_FUNCTION", enabled)` (functionControl endpoint), matching the SOC setter pattern so the integration can feature-detect by method presence.

**eg4_web_monitor**:
- Coordinator: carry `enabled` in the existing `ac_couple_soc` device-data store (same 5-minute throttled fetch, carry-forward, and post-write seeding — `note_ac_couple_soc_written` learns the bool).
- New `switch.py` entity **AC Couple** — created when a cloud client is present (not family-gated, same as the SOC numbers), writes cloud-routed in all modes, `available` only while the store holds a bool (a device that lacks the param must show unavailable, never a fake off).
- Translations + tests.

## Notes
- Distinct from the GridBOSS **port-level** `FUNC_AC_COUPLE_EN_{n}` midbox functions pylxpweb already has — this is the inverter's own smart-port AC couple function.
- No local Modbus register is pinned for this function (same evidence gap as the SOC params in pylxpweb#235) → cloud-routed writes, LOCAL-only setups don't get the entity.
[...]

Issue #352, opened 2026-07-09
AC Couple Start/End SOC %


mjstrand:
### Problem or Use Case

I have been doing a lot of testing getting my grid tied solaredge inverter ac coupled into my 12000XP v2. I have put in a 120A DPDT manual relay with 120v coil. I can control the coil with home assistant. Obviously if the grid power and the EG4 ac couple power are both live, they are at different parts of their AC cycles. Both the solaredge and the EG4 take minutes to synchronize with the ac grid before power generation starts. Yesterday the grid was down for a couple hours and solar was generating to the EG4. When the grid came back up, I mistakenly flipped the relay from the EG4 smart port to the grid without turning off the smart port breaker first.  Thankfully it did not blow up my solaredge with the out of phase power, but it did instantly stop generating and throw a fault on the solaredge. I have found that if the batteries are above the AC Couple Start and Stop SOC %, there is no power provided to the Smart Port. I would assume this would work the same with other inverters but am not sure that they have these settings.

### Proposed Solution

I am hoping to be able to create a script that would provide a safe transition of the solaredge inverter from the grid to the EG4 ac coupled smart port and back again. The EG4 monitoring web site sets ac couple start with a post to _12K_HOLD_AC_COUPLE_START_SOC and stop with _12K_HOLD_AC_COUPLE_STOP_SOC. I propose:

### Transfer to EG4 from Grid

1. If battery is less than 12%, send an alert to manually switch and abort
2. set AC Couple Start SOC to 1%
3. set AC Couple Stop SOC to 11%
4. delay 10 seconds for software to drop power to the smart port
5. power on the 120A relay transferring lines to the Smart Port with no voltage
6. delay 10 seconds to allow the solaredge to shut down entirely
7. set AC Couple Start SOC to 85% (my normal start point)
8. set AC Couple EndSOC to 95% (my normal end point)
9. that will activate the solaredge on the EG4 within 5 minutes when battery is below 85%

### Transfer to Grid from EG4

1. If battery is less than 12%, send an alert to manually switch and abort
2. set AC Couple Start SOC to 1%
3. set AC Couple Stop SOC to 11%
4. delay 20 seconds to allow the software time to drop power and solaredge to shut down entirely
5. power off the 120A relay transferring lines to the Grid with voltage
7. set AC Couple Start SOC to 85% (my normal start point)
8. set AC Couple EndSOC to 95% (my normal end point)
9. that will activate the solaredge on the grid within 5 minutes

[...]

PR #473, opened 2026-07-17
feat: AC Couple switch entity (#471)

btli:
## Summary
A per-inverter **AC Couple** switch that toggles the portal's `FUNC_AC_COUPLING_FUNCTION` — de-energizing the AC-coupled smart-port source at any battery level. Follow-up to the #352 AC Couple Start/End SOC window, suggested by @mjstrand ("drop the ac couple power at any battery level") and @ivanfmartinez, who confirmed the param on his on-grid hybrid LXP.

Cleaner than driving the SOC window for a grid↔smart-port transfer script: one independent on/off, scriptable via `switch.turn_off`/`turn_on`.

## What it does
- New **AC Couple** switch, created wherever a cloud client is present (not family-gated — the #352 evidence spans off-grid and grid-tied hardware). State from the coordinator's dedicated `ac_couple_soc` store (5-minute throttled cloud getter + carry-forward + post-write seeding); writes cloud-routed in every mode. **Unavailable — never a fake OFF —** while the store lacks a bool (device without the param, or pre-0.9.39b3 pylxpweb).
- 13 locales + README worked example (switch, plus the SOC-window alternative).

## Race + carry-forward hardening (also fixes the shipped #352 numbers)
The tri-vendor review of this PR surfaced two issues in the shared `ac_couple_soc` store that the #352 number entities already ship with; fixed at the store level so both benefit:
- **Mid-cycle write race**: a persistent per-serial, per-field write-seed registry that lives *outside* `self.data`, so a write acknowledged while a refresh cycle is in flight is not reverted when that cycle publishes a pre-write store snapshot. Each field's seed carries its own timestamp and is applied/cleared independently — a later write to one key never renews another key's seed and clobbers a legitimate external portal change an in-flight read picked up.
- **Partial-range read wiping siblings**: a `None` field on an otherwise-successful read is indistinguishable from a partial range-read failure (the SOC params' true registers are unpinned, so same-range locality can't be assumed), so a known value is carried forward per-field instead of wiped.

## Root cause / design
No inverter-level AC couple enable existed. Mirrors the #352 cloud-only architecture exactly (dedicated store, not the parameter cache — a HYBRID local refresh rebuilds the cache from local registers and would wipe cloud-seeded values). Local register path (reg 179 bit 11 candidate) deferred to #472 pending a live pin.

## Test plan
- New: entity creation gate (cloud present, not family-gated; absent in pure-local); read/availability (absent/None/corrupt → unavailable; parameter-cache wipe survived); cloud-routed writes in every mode with store seeding; failure/exception revert; missing-method (pre-0.9.39b3) explicit error. Store: per-field carry-forward both directions; the mid-cycle race (seed survives stale in-flight read; post-write read clears; independent per-field timestamps); switch→registry wiring through a real seed method.
- **2224 tests pass**; `ruff` + `mypy --strict` clean. The reviewer-flagged intermittent unit test was hardened (fragile monotonic comparison removed) and ran clean across 7 full combined-file runs.

## Risk-accepted
- **No staleness ceiling** on the carried-forward store: a device whose params permanently vanish would pin last-known values. This is the explicit #380 design decision, unchanged, and applies equally to the shipped #352 numbers — a cloud outage in HYBRID (healthy local coordinator) shows the last-known value rather than going unavailable. Consistent across all three AC-couple entities.
- **Asymmetric bool guard** (LOW): per-field carry-forward guards `enabled` with `isinstance(bool)` but not the numeric SOC fields; the read side (`number.py`) already excludes bools from numeric fields and nothing writes a bool to those keys.

## Review
Tri-vendor adversarial review (Claude + Codex GPT-5.6-sol xhigh + Gemini 3.1 Pro), 2 rounds. Every confirmed BLOCKER/HIGH/MEDIUM fixed or risk-accepted above; LOWs addressed or noted.

## ⚠️ Merge ordering
Requires **pylxpweb >= 0.9.39b3** (manifest + `tests/requirements-test.txt`). Publish [joyfulhouse/pylxpweb#237](https://github.com/joyfulhouse/pylxpweb/pull/237) to PyPI **before** merging this — until 0.9.39b3 is on PyPI, this branch's CI dependency resolution will fail.

Refs #471, #472, #352.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - **AC Couple Start/End SOC number entities** ([#352](https://github.com/joyfulhouse/eg4_web_monitor/issues/352), requested by @mjstrand): the SOC window governing the AC-coupled source on the inverter's smart port — enabled when battery SOC drops below *Start*, disabled above *End* — scriptable for safe grid/smart-port source transfers. Cloud-only (no local Modbus register): created on Cloud and Hybrid connections for all supported inverters, refreshed from the cloud on a 5-minute cadence (which also picks up portal-side edits), unavailable on devices that do not carry the parameters. A factory-disabled End threshold (`255`, "never stop") shows as unknown with a `disabled_sentinel: true` attribute. Requires pylxpweb ≥ 0.9.39b2.
+- **AC Couple switch** ([#471](https://github.com/joyfulhouse/eg4_web_monitor/issues/471), follow-up to #352 suggested by @mjstrand and @ivanfmartinez): enables/disables the inverter's AC couple function (`FUNC_AC_COUPLING_FUNCTION`) outright — de-energizing the AC-coupled source on the smart port at any battery level, without driving the Start/End SOC window. Same cloud-only architecture as the SOC pair (state from the 5-minute cloud read, writes cloud-routed in every mode, unavailable on devices that lack the function param). Requires pylxpweb ≥ 0.9.39b3.
 
 ## [3.5.1-beta.1] - 2026-07-15
 
--- a/README.md
+++ b/README.md
@@ -233,15 +233,18 @@ automation:
           entity_id: switch.18kpv_1234567890_battery_backup
 ```
 
-### AC Couple Start/End SOC (cloud connection required)
+### AC Couple controls (cloud connection required)
 
-Inverters expose **AC Couple Start SOC** and **AC Couple End SOC** number
-entities: the AC-coupled source on the smart port is enabled when battery SOC
-drops below *Start* and disabled above *End*. Scripting these lets you
-de-energize the smart port on demand — e.g. to safely transfer a grid-tied
-inverter between the grid and the smart port (#352). The parameters are not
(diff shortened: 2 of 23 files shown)


btli:
**Post-PR `/code-review` (xhigh) — one finding, fixed in `51b8f44`.**

The high-recall pass caught a correctness gap that the two-round tri-vendor review missed:

- **Seed cleared on an unobserving read.** The per-field seed-clear dropped a field's seed whenever a read *started* after the write — even a read that never *observed* the field (a partial FUNC-range failure returns `enabled=None`). Combined with per-field carry-forward pulling a race-reverted pre-write value from `self.data`, a just-written switch state could revert for one poll cycle and lose its seed protection. Fix: supersede a field's seed only when the read observed a concrete value for it (`seed.at <= now AND observed[field] is not None`); otherwise retain and apply it. A genuine external portal change still wins (the read observes the new value and clears the seed). New regression test `test_seed_survives_read_that_missed_the_field`; Codex gpt-5.6-sol verified the fix clean.

Two LOWs noted and not changed: the asymmetric `isinstance(bool)` guard on the numeric SOC fields (risk-accepted — the read side already excludes bools and nothing writes one there), and the `note_ac_couple_soc_written` early-return-without-listener edge (unreachable in practice — the switch is created from present device data).

Issue #484, opened 2026-07-25
Smart load grid always on button


brendonlobo123:
### Problem or Use Case

Looking to set the grid always on button from HA

<img width="1080" height="2424" alt="Image" src="https://github.com/user-attachments/assets/395eac49-5271-44ec-ac88-54f5eb2d5d85" />

### Proposed Solution

Expose the option

### Affected Device Model(s)

12000XP

btli:
Implemented and merged (#496) — but **please treat this as needing your confirmation before it counts as fixed.**

A **Grid Always On** switch now appears on inverters. It writes through the EG4 cloud exactly as the portal's Maintenance → Remote Set page does. It ships **disabled by default**, so you will need to enable it in the entity settings first.

## What is verified, and what is not

I confirmed the **read** side against live hardware: `FUNC_ON_GRID_ALWAYS_ON` comes back as a named parameter on an 18KPV, a GridBOSS and a FlexBOSS21, and it is present in the register range the integration actually uses to build its parameter cache. So the switch should reflect your inverter's real state.

**The write has never been tested on an inverter.** It sends the same call the portal sends, but no unit has been observed acting on it — so toggling it is the unknown part. That is exactly what would be useful from you: enable the entity, toggle it, and check whether the portal page agrees afterwards. If it does not take effect, please say so and I will dig further rather than leave it looking supported.

Leaving this issue open until you have had a chance to try it.

## Why it is cloud-only

The function lives in holding register 179, but *which bit* has never been confirmed. Guessing is not a safe failure here: the firmware acknowledges the write either way, so the integration could not detect the mistake and might silently flip an unrelated setting. If anyone can capture a raw register-179 read before and after toggling Grid Always On in the portal, that pins the bit and the local path can be added — until then the cloud is the only route known to hit the right thing.

Practical consequence: the switch is not created on Local-only connections, or on Hybrid connections with a local transport attached, because there its state could not be read at all. Where the state genuinely is unknown it reports **unavailable** rather than showing a misleading "Off".

## Not included

The same portal section has a **Smart Load Enable** toggle, which reads identically. I deliberately left it out: nobody has asked for it, its write path would be just as unverified, and it repurposes the physical port rather than adjusting behaviour within it. Happy to add it if it is useful to you — worth its own issue so the two can be confirmed independently.

brendonlobo123:
Hi Bryan,

I was able to toggle the switch off and on, and refreshing the portal
[...]

Issue #476, opened 2026-07-20
Off-Grid Mode switch non-functional in Hybrid connection mode (works fine in Cloud mode)


scottjvincent:
### Integration Version

3.5.0

### Home Assistant Version

2026.07.3

### Connection Mode

Hybrid (Local + Cloud)

### Device Model(s)

18kPV

### Describe the Bug

The Off-Grid Mode switch entity does nothing when toggled from Home Assistant while in Hybrid connection mode — no state change, no error, and toggling it either direction from HA has zero effect on the physical inverter.

### Steps to Reproduce

1 Set integration to Hybrid connection mode
2 Toggle the Off-Grid Mode switch entity from HA
3 Observe: inverter mode does not change (confirmed via EG4 monitor site/app), and the switch state in HA does not update

### Home Assistant Logs

```shell
Enabled debug logging for custom_components.eg4_web_monitor and toggled the switch — no log entries at all were generated for the attempted write, at any logging level.
```

### Screenshots

Workaround found:
Switching the integration to Cloud connection mode, the same Off-Grid Mode switch works correctly and reliably. This suggests the local Modbus write path for this entity may not be implemented yet in Hybrid mode.

Additional context:
Also noticed the Off-Grid Schedule time entities show as "unavailable" — possibly related, possibly a separate feature-detection issue.

btli:
## Root Cause Analysis

Thanks for the detailed report — the workaround observation ("works in Cloud, dead in Hybrid") pointed straight at the local register path, and we've now **hardware-proven the root cause** on an 18kPV (same model as yours).

### Root cause

The local register table (pylxpweb) maps the green/off-grid mode function (`FUNC_GREEN_EN`) to **register 110, bit 8** for the 18kPV family. The actual bit on 18kPV firmware is **bit 14**.

Live toggle test on an 18kPV (2026-07-21), toggling Off-Grid Mode via the cloud API while reading back the raw register:

```
BEFORE:   reg110 = 1056  (0000010000100000)  FUNC_GREEN_EN=False
enable green mode (cloud) → success
AFTER ON: reg110 = 17440 (0100010000100000)  FUNC_GREEN_EN=True
flipped bit: 14   ← not bit 8, where the local table points
RESTORED: reg110 = 1056
```

### Why each symptom follows

[...]

PR #501, opened 2026-08-01
fix: unavailable, not a fake OFF, when Charge Last / Share Battery state is absent (#497)

### LINKED ISSUE #497: Charge Last and Share Battery show a fake OFF when their function param is absent
Split out from the #484 review rather than swept into it.

## The shape

`EG4WorkingModeSwitch.is_on` derives state as `_parameter_data.get(param_key, False)`. When the key is **absent** — a first cloud fetch still pending, an older pylxpweb getter, or a device whose family genuinely lacks the function — that is indistinguishable from the device reporting OFF. The switch presents a confident, toggleable **Off** for a control whose state is unknown.

#471 (AC Couple) solved this with an `available` override, whose docstring names the case exactly:

> Absent state — first cloud fetch pending, a pre-0.9.39b3 pylxpweb getter, or a device whose family genuinely lacks the function param — must show unavailable, never a fake OFF.

#484 (Grid Always On) hit the same problem, because dropping its family gate made "device lacks this param" a live case rather than a contradiction. It was fixed there with an opt-in `requires_known_state` flag on the mode definition: when set, an absent state key reads `available=False` / `is_on=None`.

## What is still exposed

**`FUNC_CHARGE_LAST` (Charge Last)** and **`FUNC_BAT_SHARED` (Share Battery)** are ungated the same way and carry the identical exposure. Neither sets `requires_known_state`.

The flag deliberately was not applied to the shared base for all eleven working modes: that would flip every working-mode switch from OFF to unavailable during the pre-first-parameter-read window — a user-visible change to long-standing behaviour, and well outside #484's scope. So the two remaining cases are called out here instead of being changed silently.

## What to decide

Whether these two should opt in. Arguments each way:

- **For**: they are ungated by family, so an absent param is reachable, and a toggleable Off for an unknown state is exactly the failure #471 identified.
- **Against**: unlike Grid Always On, both have local register mappings, so on LOCAL and HYBRID their state comes from a register read rather than a cloud param — the absent-key window is narrower and mostly limited to the pre-first-read period, where flipping to unavailable is a visible behaviour change for existing users.

If they do opt in, the parity test added in #484 (driving two switches off one coordinator and asserting they agree) is the pattern to extend.

Low urgency — no user has reported being misled by it. Filed so the inconsistency is recorded rather than rediscovered.

btli:
## The problem

`EG4WorkingModeSwitch.is_on` reads `_parameter_data.get(param_key, False)`. When the key is **absent**, that collapses into a confident, toggleable **Off** — indistinguishable from the device actually reporting the function disabled.

**Charge Last** (`FUNC_CHARGE_LAST`) and **Share Battery** (`FUNC_BAT_SHARED`) are ungated by family, so "this device does not report the function" is a live case rather than a contradiction. That is exactly the exposure #471 identified on AC Couple and #484 fixed on Grid Always On. Both were named in #484 as the remaining cases and split out to #497 rather than swept in silently.

Both now set `requires_known_state`, the opt-in flag #484 added: absent key → `available=False` / `is_on=None`.

## LOCAL/HYBRID trace (done before flipping the flag)

The argument *against* opting in was that these two, unlike Grid Always On, have local register mappings — so opting in might strand them as unavailable where a cloud-only mode would not. Traced end to end; it does not:

- `_fetch_parameters` reads holding registers in two groups, `(0, 125)` and `(125, 125)`. Register 110 is in the first.
- `read_named_parameters` decodes a bit-field register by enumerating **every** param key in that register's list, so a successful read sets both keys as bools **whatever the bits hold**. Presence is not value-dependent.
- The reg-110 list is shared by every family — the `EG4_OFFGRID` branch has returned the base table since #476 unified the layout. Verified at runtime against the installed pylxpweb for `None`, `EG4_OFFGRID`, `EG4_HYBRID`: bit 3 → `FUNC_BAT_SHARED`, bit 4 → `FUNC_CHARGE_LAST` on all three.
- The #282 sticky carry-forward keeps the keys once seen, so a routine partial read cannot blank them.

The CLOUD path was checked empirically too, because there the key comes from the portal read rather than a register — an omission there would be a **permanent** unavailable, not a startup-window one. A read-only probe of plant 19147 returns both keys on all three devices (18kPV `4512670118`, GridBOSS `4524850115`, FlexBOSS21 `52842P0581`), ~193 keys each, both `False`.

## Known behavior change (accepted)

On every connection type, these two switches now read **unavailable rather than Off** during the pre-first-parameter-read window at startup. That is the deliberate trade-off: a brief unavailable beats a confident wrong Off. Stated in the changelog entry.

## Tests

`TestKnownStateWorkingModeParity` extends the #484 parity pattern — it drives both switches, plus AC Couple and Grid Always On, off **one** param-less coordinator and asserts they agree. That divergence going unnoticed for a release is the whole reason this issue exists.

Per mode (parametrized over both): absent → unavailable/`None`; present `False` → a real, toggleable Off; present `True` → on; LOCAL register-sourced state → available; optimistic mid-write → available; and a wholesale partial read that drops the key → back to unknown rather than Off.

Two existing tests encoded the old behavior and were updated rather than deleted:
- `test_other_working_modes_keep_the_historical_off_default` now derives its list from the modes that did *not* opt in, with a guard so it cannot silently empty out if every mode opts in later.
- `TestChargeLastSwitch::test_is_on_false_default` asserted the fake Off directly; renamed to `test_missing_param_is_unknown_not_off` and inverted to the new contract.

**Mutation-verified:** dropping the flag from `charge_last_mode` fails 4 tests, from `share_battery_mode` fails 3 — the parity test catches both.

## Validation

- `2348 passed, 3 skipped` (switch suite alone: 226 passed)
- `ruff check` + `ruff format`: clean
- `mypy`: 57 errors **both before and after** this diff — pre-existing in this local env, 35 of them `import-not-found` for `pylxpweb`, which the hook's mypy cannot resolve. The commit therefore used `--no-verify`; this diff touches only a dict literal entry, docstrings, and tests, and changes the error count by zero.

## Scrutiny points

- **The setup probe reads the base table, not the family-resolved one.** `_local_params_can_carry` consults `REGISTER_TO_PARAM_KEYS` directly rather than `get_register_to_param_mapping(family)`. Harmless today because all families share reg 110, but it is a blind spot if a family ever diverges again. Not changed here — out of scope, worth its own issue.
- **Share Battery is `enabled_default: False`**, so its visible blast radius is limited to users who explicitly enabled it. Charge Last is enabled by default and is the one users will actually notice at startup.
- **Cloud key presence is evidenced on three devices from one plant**, not proven for every family. If a family's portal read omits either key, that switch becomes permanently unavailable there rather than showing a wrong Off — which is the intended semantic, but it would be a visible change for those users.

Fixes #497


btli:
## Adversarial review round (Codex gpt-5.6-sol, high)

**Verdict: no P1/P2. One P3, confirmed and fixed in b06eaff.**

**P3 — the LOCAL test bypassed the behavior it claimed to verify.** `test_local_register_sourced_state_stays_available` hand-seeded `parameters={param: False}`, so it never decoded register 110, never selected a family map, and never touched `_fetch_parameters`. It passed whether or not a local read can actually supply these keys, and it survived deleting `requires_known_state` entirely.

That matters more than its P3 label suggests, because it was the *only* test standing behind the single assumption this whole change rests on: opting in is safe **only** because a bit-field read populates both keys regardless of the bits' values. A vacuous test there is a fake guarantee.

Now decoded for real — a `BaseTransport` stub returns a raw register 110, pylxpweb's own `read_named_parameters` produces the dict, and it is handed to the switches exactly as `_fetch_parameters` hands it to the coordinator cache.

**Both bit polarities are load-bearing, not decorative.** Verified by mutating pylxpweb:

| Mutation | `bits-clear` (0x0000) | `bits-set` (0x0018) |
|---|---|---|
| Drop `FUNC_CHARGE_LAST` from the reg-110 list | catches | catches |
| Emit only SET bits (register decodes to a raw int under every name) | **passes** — still reads `False` | catches (raw `24`, `24 == 1` → `False` vs expected `True`) |

A single all-clear case would have missed the second mutation entirely.

The per-family half of this contract was already covered elsewhere: `test_register_110_contract_holds_for_every_family` pins `FUNC_CHARGE_LAST` to reg 110 bit 4 and `FUNC_BAT_SHARED` to bit 3 for every `InverterFamily`. What was missing, and is added here, is the decode *behavior* that turns those positions into present keys.

Codex's other probes came back clean and I confirmed them: no unsafe `None` consumer (optimistic retention explicitly handles `bool | None`; `extra_state_attributes` and write-seeding are safe), neither switch uses restore-state, registry defaults and Share Battery's disabled-by-default status unchanged, and the modified existing tests did not lose substantive coverage — the dynamic unflagged-mode list is strictly broader than the previous hardcoded three.

CI: all checks pass, including **Platinum - Strict Typing**, which also confirms the local mypy noise mentioned in the PR body is purely environmental.

Issue #255, opened 2026-06-15
Harden Quick Charge against rapid/concurrent register writes + reads


btli:
## Summary

Harden the Quick Charge register access against **rapid/concurrent writes, modifications, and reads**. Reported by @ivanfmartinez on #251 while testing `v3.4.0-beta.13`:

> Only one time I have tried to change Duration to 3 and it kept on 59, probably a concurrency with the update cycle (these don't appear to be a big problem, and can occur on any entity), tried again and worked.

Low impact (a retry works, and it's a general HA-entity class of problem), but worth hardening for robustness. Not a regression.

## Why it happens

Quick Charge touches the same holding/input registers from several independent paths that contend on one Modbus transport:

- The **coordinator poll** reads regs 233 + 234 (and INPUT 210) every cycle (status throttled ~30s).
- **`is_quick_charge_active_live`** does an on-demand fresh read of reg 233 when the Duration number is set.
- The **Duration write** sets HOLD 234 (`write_named_parameter`) while a charge is active.
- The **switch** does a read-modify-write of HOLD 233 bit 0 (enable/disable).

Because the `Quick Charge Duration` number now **mirrors the live register** (no optimistic state), a write that lands at nearly the same moment as a poll read can be "overwritten" on screen by the poll's value (the reporter's 3-vs-59), or a write can be lost to bus contention.

## Edge cases to address

1. **Write-vs-poll race** — user writes HOLD 234 = N while a poll reads HOLD 234 = (live countdown); the number snaps back to the polled value, looking like the write was lost.
2. **TOCTOU on the active gate** — `is_quick_charge_active_live` returns True, the charge auto-expires, then the HOLD 234 write is rejected by the firmware (surfaces as an error). Currently accepted as rare.
3. **Transport flap during a write (HYBRID)** — if `has_local_transport` is briefly False (transport reconnecting), a Duration set is silently routed to the cloud preference branch instead of writing the register (no false success, but a silent no-op on the display). Raised in adversarial review.
4. **RMW interleave on reg 233** — the enable-bit read-modify-write is op-lock-serialized inside pylxpweb, but there's no single lock spanning the integration's multiple entry points (poll read + on-demand read + write).

## Hardening directions to investigate

- **Optimistic state** on the Duration number: show the just-written value immediately and reconcile on the next poll (instead of snapping to the last polled register value).
- **Immediate targeted re-read** after a write, rather than waiting for the next poll cycle, so the display converges quickly.
[...]

PR #503, opened 2026-08-01
feat: AC Couple local Modbus path via holding register 179 bit 11 (#472)

### LINKED ISSUE #472: AC Couple local register path: pin reg 179 bit 11 (FUNC_AC_COUPLING_FUNCTION)

btli:
Fixes #472.

## What

Gives the AC Couple switch (#471) a LOCAL Modbus path. `FUNC_AC_COUPLING_FUNCTION` is now mapped to holding register 179 bit 11 in pylxpweb, so the switch can ride the ordinary parameter cache and the local named-write path like every other reg-179 function (Export PV Only, Grid Peak Shaving).

| Mode | Before (#471) | After |
|---|---|---|
| CLOUD | store read + `set_inverter_ac_couple_enabled` write | **unchanged** |
| HYBRID | store read + cloud write | local-first read (reg 179 bit 11) + local write with cloud function-control fallback |
| LOCAL | **entity not created at all** | entity created; reads and writes over Modbus with no cloud client |

The cloud `ac_couple_soc` store stays: the Start/End SOC pair beside it (#352) still has no pinned register, and the store remains the read fallback whenever the parameter cache carries no value.

Absent state still reads **unavailable, never a fake OFF** (#497 pattern), from either source.

A successful local write deliberately does **not** seed the cloud store. While bit 11 is unpinned that store is an *independent* observation of the same function, and seeding it would mask a disagreement between the bit we wrote locally and what the portal reports — which is exactly the signal #472's lockstep capture is after.

## Evidence for bit 11 — stated precisely

Bit 11 is **NOT pinned by this project's raw↔named lockstep toggle**. It ships on lineage inference:

1. The Luxpower Modbus doc places AC coupling at reg 179 bit 11.
2. `ant0nkr/luxpower-ha-integration` (`constants/hold_registers.py`, `H_FUNCTION_ENABLE_4 = 179`) documents `Bit 11: uFunctionEn2.ubACcoupling (0=Disable, 1=Enable)`. Its full 16-bit reg-179 layout matches this project's canonical `inverter_holding.py` table bit-for-bit, and four bits of that shared layout are independently hardware-proven on EG4 hardware: bit 3 (2026-06-12 raw-verified cloud toggle on a FlexBOSS21 and an 18kPV), bit 7 (live toggle), bits 9/10 (2026-02-18 SOC-vs-Voltage). No bit of the reg-179 layout has been found wrong.
3. #471's reporter (@ivanfmartinez) has driven this control through that mapping on his LXP hardware before, and his live named-read correlation agrees: his AC-couple-enabled LXPUS810K reads the cloud param `True`, a disabled SNA12K-US probe reads `False`.

**What is missing:** the strict raw↔named lockstep toggle (flip the named param, read raw reg 179 before/after, verify the XOR is exactly bit 11, flip back). This is the same standing on which the #476 off-grid green-mode bit shipped.

Writes go through pylxpweb's sibling-preserving read-modify-write, and the entity's post-write refresh re-reads the register, so a write that **fails to land** is visible. To be explicit about what that does *not* buy: a write to the **wrong** bit would still be ACKed by the firmware, so readback-verify cannot rule that out — only the toggle capture can. #472 asks for it.

**Guards:** `_CONTROL_REGISTER_CONTRACT` pins the name to (179, 11), and a new `test_register_179_contract_holds_for_every_family` mirrors the reg-110 guard added for #476 so a family override can never silently move the bit.

## Freshness trade-off (deliberate, please weigh)

In HYBRID the parameter cache refreshes on the parameter tier (`parameter_refresh_interval`, default **60 min**) while the cloud AC-couple store refreshes every **5 min**. Preferring the local value therefore surfaces a *portal-side* toggle later than #471 did. That is what local-first means and matches every other reg-179 switch; changes made from HA are unaffected (the write path updates the cache in place). Called out in the class docstring. Say the word if you'd rather HYBRID keep reading the 5-minute store — that inverts the read preference and requires re-adding store seeding on local writes.

## Known gap: no capability probe on pure LOCAL

In #471 the cloud getter's `None` doubled as the capability probe — a device whose family lacks `FUNC_AC_COUPLING_FUNCTION` simply did not report it, and the switch went unavailable. Reg 179 bit 11 has no such tell: it decodes to a bool on any device that answers the register, so it is a **state source, never an existence proof**. A **pure-LOCAL** install therefore shows the switch on every control-capable inverter, reading OFF where the hardware has no AC-coupled input.

There is no local capability signal to gate on (pylxpweb's `InverterFeatures` has no AC-couple or smart-port flag), and a model-string gate would repeat the #259 mistake. Documented at the creation gate and asserted by a test so the gap cannot silently widen or close unnoticed.

**HYBRID does NOT share this gap** — but an earlier revision of this PR only *claimed* it didn't. Local precedence overrode the absent cloud probe, so HYBRID published the same phantom switch. Availability (and `is_on`, via one shared `_capability_known` notion) now consults the cloud store's tri-state wherever a cloud client exists.

## Post-write verification

A local write forces `async_refresh_device_parameters` before returning, re-reading register 179. This too was initially only *claimed*: the shared local-write envelope mutates the cached value and runs a coordinator DATA refresh, but parameters live on their own tier, so the register was not actually re-read until the next scheduled parameter cycle (default 60 min). For a lineage-inferred bit that readback is the whole point. It goes through the same public method the cloud path uses, so pylxpweb's #282 partial-read carry-forward and retry floor apply unchanged; it never raises, and a failed verification warns rather than failing a command the device already acknowledged.

To be precise about what this buys: it makes a write that **failed to land** visible within the cycle. A write to the **wrong bit** would still ACK and read back as written — only the #472 toggle capture rules that out.

## Mutation verification

Both fixes above were reverted independently to confirm the new tests actually catch the defects they describe — **5 distinct tests, not 5 per fix**:

| Mutation | New tests that fail |
|---|---|
| Revert the capability gate to local-first | **2** (`test_hybrid_unavailable_when_cloud_never_reported_the_param`, `test_hybrid_unavailable_before_the_first_cloud_fetch`) |
| Remove the post-write readback call | **3** (`test_local_write_forces_a_verifying_parameter_reread`, `test_failed_reread_warns_but_does_not_fail_the_write`, `test_local_only_write_also_rereads`) |

All 43 AC-Couple tests pass with the fixes in place. The pylxpweb side carries the same treatment: reverting `set_ac_couple` to the raw read-modify-write makes the reg-179 atomicity test fail with `register_value == 2560` — bit 7 dropped.

## Dependency — this PR is expected RED until pylxpweb ships

Requires the companion pylxpweb PR (targeting **0.9.39b6**). The manifest pin is **deliberately not bumped** here. Until that release lands, 6 register-contract tests fail on purpose — the same "DELIBERATELY RED against older pylxpweb" coupling bit 3 had against 0.9.36b6. Verified locally against the pylxpweb branch:

- installed pylxpweb (0.9.39b5), i.e. what CI runs: `2385 passed, 6 failed` — all 6 are the intended contract-pin failures
- pylxpweb **main** on `PYTHONPATH` (the mapping is merged there now): **`2394 passed, 3 skipped, 0 failed`**

Older installs that lack the mapping keep the exact pre-#472 cloud-only behavior via the `_local_params_can_carry` probe (setup gate + per-entity guard), and there is a test for that.

## Validation

- `pytest tests/`: 2394 passed, 3 skipped (against merged pylxpweb main); 43 AC-Couple switch tests, up from 21
- `ruff check` / `ruff format --check`: clean
- `mypy --config-file tests/mypy.ini custom_components/eg4_web_monitor/`: `Success: no issues found in 40 source files`
- The pre-commit hook's mypy runs under a python without HA/pylxpweb and reports 57 errors — **unmodified `origin/main` reports the same 57**, zero new

## Not covered

No hardware run. The switch has not been exercised on a live LOCAL or HYBRID install; #472 stays open for the toggle capture.


btli:
CI is red exactly as predicted in the description, and only there: **6 failed, 2345 passed, 3 skipped** ([run](https://github.com/joyfulhouse/eg4_web_monitor/actions/runs/30711507863)). All six are the deliberate cross-repo pin:

```
test_control_params_resolve_to_documented_registers
test_register_179_contract_holds_for_every_family[EG4_OFFGRID|EG4_HYBRID|LXP|UNKNOWN|no-family]
  AssertionError: FUNC_AC_COUPLING_FUNCTION absent from register 179 — writes fail
```

That is the contract harness reporting that the installed pylxpweb (0.9.39b5) does not yet carry the mapping. Every other check on the PR passes. With joyfulhouse/pylxpweb#259 on `PYTHONPATH` the same suite is **2351 passed, 3 skipped, 0 failed**.

pylxpweb#259 is CI-green (unit + integration + lint/type + claude-review). Merge and release it as 0.9.39b6, then bump this repo's manifest pin at the release cut and this goes green with no code change.

btli:
Fix round pushed (`45b8edc`). Both P2s were real, and both were the same failure mode: **a comment promising a protection the code did not implement.** Worth naming that pattern — it is the second time on this pair (the pylxpweb #259 round had the old table entry's "confirmed on 12000XP" annotation).

**P2-1 — phantom switch in HYBRID.** Confirmed exactly as described. Reg 179 bit 11 is a state source, never an existence proof: it decodes to a bool on any device that answers the register, so local precedence produced `available=True` / `is_on=False` on hardware with no AC-coupled input, while the comment claimed the store was protecting it. Availability now consults the cloud store's tri-state wherever a cloud client exists, HYBRID included.

One thing I extended beyond the letter of the ask: `is_on` shares the same `_capability_known` notion. Gating only `available` would leave an unsupported device *unavailable-but-confidently-OFF* — invisible in the UI, but a lie to anything reading state directly, and exactly the #497 shape. The two now cannot disagree.

Pure LOCAL keeps the documented gap, and it is now **asserted by a test** so it cannot silently widen or close unnoticed.

**P2-2 — verification never happened.** Confirmed: the shared envelope mutates the cached value and runs a coordinator DATA refresh, but parameters are on their own tier, so reg 179 was not re-read until the next scheduled cycle (default 60 min) — the test fixture's parameter-refresh mock was indeed never invoked. Implemented the real reread rather than taking the honest-comment alternative: the local write path now forces `async_refresh_device_parameters`, the same public method the cloud path already uses, so pylxpweb's #282 partial-read carry-forward and retry floor apply unchanged. It never raises; a False return warns and leaves the commanded value showing, because a failed verification must not fail a command the device already acknowledged. Tests assert the await, the pure-LOCAL path, and the warning.

**Both fixes are mutation-verified**, and the two mutations were re-run INDEPENDENTLY to pin the split (an earlier phrasing here read as "5 each" — it is 5 total):

| Mutation | New tests that fail |
|---|---|
| Revert the capability gate | **2** — `test_hybrid_unavailable_when_cloud_never_reported_the_param`, `test_hybrid_unavailable_before_the_first_cloud_fetch` |
| Remove the readback call | **3** — `test_local_write_forces_a_verifying_parameter_reread`, `test_failed_reread_warns_but_does_not_fail_the_write`, `test_local_only_write_also_rereads` |

2 + 3 = 5 distinct tests; all 36 pass with both fixes in place.

Also corrected the PR body, which had repeated the same false claims (HYBRID probe, post-write verification), and stated plainly what the readback does and does not buy: it catches a write that failed to land, not a write to the wrong bit.

Gates: **2358 passed / 3 skipped** (+7) against the pylxpweb branch, ruff clean, mypy clean on 40 files. CI will show the same 6 deliberate contract reds.

btli:
Round 3 pushed (`4d8844d`). Items 2 and 3 fixed; item 1 dropped per the merge plan (canonical fix lands in #504's shared `_fetch_cloud_param_store` refactor — patching the pre-refactor `_fetch_ac_couple_soc` here would only create a conflict).

**Item 2 — unbounded unverified window.** Confirmed: cache seeded, optimistic cleared, False branch only warned, so persistent read failures would show the commanded value for up to the parameter interval with nothing ever checking it. The failure path now re-arms the #282 per-device retry via a new `note_parameter_verification_pending()`, so the read retries past the ~2-minute attempt floor instead of at the next hourly window.

Deliberately **not** an immediate forced read — bypassing the floor is how a device with a failing transport gets hammered every cycle. The commanded value staying visible for those couple of minutes is the acceptable part; unbounded is not, exactly as you framed it.

**Item 3 — double reread.** Confirmed and reproduced (`await_count == 2`). `_execute_local_with_fallback` gains an optional `after_local_write` hook invoked **only when the local route succeeded**, so the cloud fallback keeps its envelope's single refresh and the local path does its own. One write = one read on both routes. The kwarg defaults to None, so no other caller's behavior changes.

**Verification.** Both mutation-verified: removing the re-arm fails the retry test; making the readback unconditional again reproduces the two-read count precisely. The re-arm also has a test that binds the **real** coordinator method over a **real** set — the mock coordinator's auto-mock records the call and does nothing, so a wrong attribute or wrong serial would otherwise pass silently.

Gates: **2363 passed / 3 skipped** (+5) against the pylxpweb branch, ruff clean, mypy clean on 40 files. Same 6 deliberate contract reds expected in CI.

btli:
Rebased onto eg4 main and force-pushed: `a0de02c`. **`BEHIND` → `UNSTABLE`** (i.e. up to date; only the expected contract reds remain). Flagging why the rebase was needed: eg4's branch protection is `strict: true`, so a BEHIND branch cannot merge no matter how green it is — main had moved on by 8 commits (#501/#497 and #507/#506).

Clean textually. I checked for a **semantic** collision rather than trusting that, since #501 landed the attach-recovery re-arm my item-2 fix was modeled on: `_param_retry_pending` is the canonical re-arm set on main, #501 feeds it by clearing the ATTEMPT stamp, and my `note_parameter_verification_pending()` adds to the same set. Consistent, not a parallel mechanism — and there was no entity-facing helper before, which is why the new one is warranted.

Also re-validated against **merged pylxpweb main** (a620a6a for #259, plus #257 on top) rather than my old branch: **2391 passed, 3 skipped, 0 failed**, ruff clean, mypy clean on 40 files.

CI: **6 failed, 2385 passed** — the same six contract reds, unchanged by the rebase. They resolve when 0.9.39b6 hits PyPI and the manifest pin bumps; the mapping itself is already on pylxpweb main, so with it on `PYTHONPATH` the suite is fully green.

Items 2 and 3 from the last round are in `4d8844d` (now rebased), both mutation-verified. Ready to merge behind the release.

btli:
Round 4 pushed (`bc77085`) — and the CI picture changed underneath us, in a good way.

## The P2 was real, and worse than "inert"

Verified the chain in code before touching it: `_param_retry_pending` is read/written **only** in `coordinator_local.py`, while HYBRID routes `_async_update_hybrid_data` → `_async_update_http_data`, whose hourly hook (coordinator_http.py:439) goes through the mode-agnostic `_should_refresh_parameters()` and never looks at that set. So last round's re-arm closed the unbounded window in LOCAL and left it open in **HYBRID — the deployment mode this repo runs in production.**

`note_parameter_verification_pending` now clears `_last_parameter_refresh` as well as adding to the set — both paths, both still floored by `_last_parameter_attempt`. Deliberately still no forced immediate read: bypassing the floor is how a device with a failing transport gets hammered every cycle. The targeted LOCAL set stays, since it re-reads one device rather than all.

**P3-1:** the shared `after_local_write` hook now isolates a raising callback — it runs inside `local_write()`, so an exception there was indistinguishable from a failed local WRITE and would have re-written via cloud a command the device already accepted.

All three mutation-verified: dropping the stamp clear reproduces the HYBRID inertness; removing the hook isolation produces the redundant cloud write; a floor test pins that the re-arm does not bypass the rate limit.

**P3-2:** body counts corrected to measured values.

## The 6 contract reds are GONE — pylxpweb 0.9.39b6 is on PyPI

CI now installs b6 (it carries #259's bit-11 mapping), so **every register-contract test passes**. The deliberate cross-repo red this PR carried is resolved without a code change.

## One failure remains, and it is NOT from this PR

```
FAILED tests/test_coordinator_local.py::TestCanonicalAbsentPredicateInRRMerge::test_degraded_row_dropped_on_older_pylxpweb
AssertionError: pylxpweb now ships is_absent(); this test's premise needs revisiting
```

That is #507/#506's own tripwire firing **exactly as designed**. Its docstring says "CI runs against a pylxpweb that has no is_absent(), so this is the path CI actually exercises today" — and b6 ships `is_absent()` (added by pylxpweb #255, `transports/data.py:1118`), so the premise is now false.

**It is main-wide, not mine.** I checked out unmodified `origin/main` into a scratch worktree and ran that single test against released b6: it fails there too. Every open eg4 PR will show it until the test is updated to construct its own predicate-less stub instead of depending on the installed pylxpweb.

I have not touched it — it belongs to the #506 work and the fix is a judgement about that test's intent, not mine to make unilaterally. Happy to take it if you'd rather not context-switch whoever owns it.

Gates on this branch against released b6: **2393 passed, 3 skipped, 1 failed** (that tripwire only). Ruff clean, mypy clean on 40 files.

btli:
Rebased onto current main and force-pushed: `f7f360b`. **FULLY GREEN — 35/35 checks pass, `mergeStateStatus: CLEAN`.** All four Tier Summaries (Bronze, Silver, Gold, Platinum) included. First time this PR has had zero reds of any kind.

What cleared it, both from main rather than from this branch:

- **#508's pin bump** to `pylxpweb>=0.9.39b6` — b6 carries #259's bit-11 mapping, so the six register-contract tests this PR deliberately carried red now pass.
- **`6fca2b6`** — the `is_absent()` tripwire, now simulating the older pylxpweb via `monkeypatch.delattr` instead of asserting the environment lacks it. My earlier "it fails on main too" check was run against the `origin/main` I had fetched at `c94fe38`, which predated that merge by four minutes. The finding was right; the "still broken" part was stale.

One thing I did change on this branch: five comments still called 0.9.39b6 "planned" / "not yet cut", and the contract entry still described itself as DELIBERATELY RED. All true when written, none true now that the pin requires b6 — corrected in `f7f360b`. The runtime guard is unchanged and so is its rationale: it probes the register map rather than trusting a version string, so a pin that somehow lacks the mapping degrades to the cloud-only path instead of writing a name the transport cannot resolve.

Local against released b6: **2394 passed, 3 skipped, 0 failed.** Ruff clean, mypy clean on 40 files.

Ready to merge.

Issue #499, opened 2026-07-31
Missing smart load options


brendonlobo123:
### Problem or Use Case

Need options for smart load exposed. 

<img width="1080" height="2424" alt="Image" src="https://github.com/user-attachments/assets/016013cb-cb67-402a-9260-d9094d4918cf" />

### Proposed Solution

Expose options

### Affected Device Model(s)

12000XP

btli:
Implemented and merged (#504, with the pylxpweb side in joyfulhouse/pylxpweb#257) — ships in **v3.5.1-beta.5**. All six controls from that panel are now there alongside the Grid Always On you confirmed for #484: a **Smart Load** enable switch, **Start/End SOC**, **Start PV Power (kW)**, and **Start/End Voltage** numbers.

A read-only check against my own account found all six settings on an 18kPV and a FlexBOSS21, so the values you see will be the real ones the portal shows. Both the SOC pair and the voltage pair are offered and writable — your screenshot greys the voltage pair out while SOC mode is active, and nothing I could find in the cloud data says which mode a unit is in, so I did not want to hide half of them on a guess.

**What would help from you is the same thing as last time: the writes.** The read side is verified; the write side has only been reasoned about, never observed on an inverter. The enable switch sends the identical kind of call you already confirmed works for Grid Always On, and the five thresholds send the portal's own named parameter writes. So: enable the entities (they ship **disabled by default** — Settings → Devices & Services → your inverter → entity list), change one, and check whether the portal page agrees afterwards. If any of them does not take effect, say so and I will dig rather than leave it looking supported.

Two things worth knowing: they are **cloud-only** (none of these has a confirmed local register, so they are not created on Local-only connections, and on Cloud and Hybrid alike the write goes through the EG4 cloud exactly as the portal does), and where a value genuinely cannot be read they report **unavailable** rather than showing 0 — that is not hypothetical, since a GridBOSS answers the enable function but carries none of the five thresholds.

Leaving this open until you have had a chance to try them.

btli:
@brendonlobo123 The Smart Load panel is released in **[v3.5.1-beta.5](https://github.com/joyfulhouse/eg4_web_monitor/releases/tag/v3.5.1-beta.5)**: the **Smart Load** enable switch plus **Start/End SOC**, **Start PV Power**, and **Start/End Voltage**, alongside the Grid Always On switch from beta.4.

[...]

PR #504, opened 2026-08-01
feat: expose the rest of the Smart Load panel (#499)

btli:
Addresses #499 (stays open for write-path hardware validation)

## What

The portal's *Maintenance → Remote Set → Smart Load Port → **Smart Load*** tab carries seven controls. #484 shipped one of them (Grid Always On); @brendonlobo123 asked for the other six on a 12000XP. This adds all six:

| Entity | Cloud parameter |
|---|---|
| `switch` **Smart Load** | `FUNC_SMART_LOAD_ENABLE` |
| `number` **Smart Load Start SOC** (%) | `_12K_HOLD_SMART_LOAD_START_SOC` |
| `number` **Smart Load End SOC** (%) | `_12K_HOLD_SMART_LOAD_END_SOC` |
| `number` **Smart Load Start PV Power** (kW) | `_12K_HOLD_START_PV_POWER` |
| `number` **Smart Load Start Voltage** (V) | `_12K_HOLD_SMART_LOAD_START_VOLT` |
| `number` **Smart Load End Voltage** (V) | `_12K_HOLD_SMART_LOAD_END_VOLT` |

All disabled by default. The SOC pair and the voltage pair *appear* to be alternatives — the reporter's screenshot greys the voltage pair out under SOC mode — but which pair an inverter acts on is **untested here**, and no mode parameter was found to gate on, so both are exposed and writable rather than one being hidden on inference.

## Evidence

Read-only cloud probe against the maintainer's own account, 2026-08-01, plant 19147, through `read_device_parameters_ranges` — the same call that builds the cloud parameter cache. On **both** an 18kPV (`4512670118`) and a FlexBOSS21 (`52842P0581`), all six keys came back, all in the `127+127` range, none in `0+127` or `254+127`:

```
FUNC_SMART_LOAD_ENABLE          = False
_12K_HOLD_SMART_LOAD_START_SOC  = '90'
_12K_HOLD_SMART_LOAD_END_SOC    = '60'
_12K_HOLD_START_PV_POWER        = '0.5'   <- kW, 0.1 resolution
_12K_HOLD_SMART_LOAD_START_VOLT = '54'    <- volts, not decivolts
_12K_HOLD_SMART_LOAD_END_VOLT   = '48'
```

The **GridBOSS** (`4524850115`) returned `FUNC_SMART_LOAD_ENABLE` and **none** of the five holdParams — it carries the per-port `MIDBOX_HOLD_SL_*` family instead.

**To be precise about what that evidence is and is not** (corrected after review — the code was always right, the earlier wording implied a GridBOSS branch that does not exist): **a GridBOSS never gets any of these entities.** Both the switch and the five numbers are created inside the `device_type == "inverter"` block, so it is not reached. What the GridBOSS probe establishes is that **the cloud really does answer with the function param present and all five holdParams missing** — the partial shape is real, not hypothetical. The case it guards is therefore an **inverter** whose portal read comes back in that same shape: the switch is available, the five numbers are unavailable. Wired as availability, not as a model list, so every entity goes **unavailable** on an absent field rather than rendering a confident `0`.

## ⚠️ The write path is unverified on hardware

The read side is verified. Writing sends the same named cloud calls the portal sends, but no inverter has been observed acting on them — the same position #484 shipped in. #499 stays open pending @brendonlobo123's confirmation.

## Design

**Cloud-routed only, in every mode**, following #352/#471 rather than #484's working-mode route. The five holdParams have no pinned local register at all, and the parameter cache cannot serve them: with a local transport attached pylxpweb rebuilds `inverter.parameters` from local registers alone and wipes anything cloud-seeded there (PR #380 review P1). They therefore need the dedicated device-data store — and putting the enable switch in that same store buys it HYBRID coverage that #484's parameter-cache route has to suppress. Register 179 membership for the function param is confirmed; the **bit** is not, and a wrong bit is ACKed by the firmware so no fallback fires and readback-verify cannot catch it (#476's reg-110 lesson).

**No second copy of the store machinery.** The `ac_couple_soc` machinery is now spec-driven (`CloudParamStoreSpec`) and both stores share it: throttled 5-minute getter reads, carry-forward on throttle/failure, per-field partial-read protection, and the write-seed registry that survives a `self.data` swap. Each store keeps its **own** throttle key and its **own** seed registry — both have a `start_soc` field, so a shared registry would let one overwrite the other. The AC-couple entry points survive as named bindings; their behavior is unchanged.

**Not family-gated**, matching #471/#484: the params answered on hybrid *and* off-grid hardware, and no family has evidence of absence.

The five holdParams and `FUNC_SMART_LOAD_ENABLE` are contracted in the register harness's cloud-only allowlist, which goes **STALE** the moment pylxpweb pins a register or bit for any of them.

## Dependency

Requires **pylxpweb ≥ 0.9.39b6** (joyfulhouse/pylxpweb#257) for the named `get`/`set_inverter_smart_load_*` methods. **The manifest pin is deliberately not bumped here** — that belongs to the release cut, once the pylxpweb release is on PyPI (bumping it now would break CI until b6 exists). Without those methods the entities stay unavailable and a write reports the required version instead of raising `AttributeError` (covered by tests).

### ⚠️ One test is DELIBERATELY RED until 0.9.39b6 ships

`test_smart_load_methods_exist_in_installed_pylxpweb` resolves the seven pylxpweb methods against the **installed** package and fails when they are absent. It exists because every other test of this feature mocks the control endpoint — so the whole suite passed green against a pylxpweb that has none of them, which is precisely the state that would ship seven permanently-unavailable entities. Red here means "the dependency is not satisfied yet", not "the integration is broken". It goes green when b6 is released and pinned, mirroring the #472 contract pin. A companion test anchors that list to the real call sites, so a rename cannot leave the pin passing against methods nothing uses.

## Validation

- **2398 tests pass** (63 new: 19 switch, 31 number, 13 coordinator-store), 3 skipped
- `ruff check` + `ruff format` clean
- `mypy --config-file tests/mypy.ini` clean under the venv

Committed with `--no-verify`: the pre-commit mypy hook runs the **global** interpreter, which cannot see pylxpweb and reports the same **57** import/decorator errors on unmodified `main` — verified identical, and this change adds none.

## Scrutiny points

- **Where the numbers' bounds came from.** SOC is 0-100. The voltage and kW ceilings (0-100 each) are *defensive sanity bounds, not firmware-confirmed maxima* — the wire carries portal units at 0.1 resolution with no documented ceiling, and a too-narrow range would render a legal portal value as unknown. The sibling `_12K_HOLD_AC_COUPLE_END_VOLT` reads `80` on our own units, which is why the voltage range is not clamped to a 48 V bank's working span.
- **`_12K_HOLD_START_PV_POWER` has no `SMART_LOAD` in its name.** Its association with this panel comes from the reporter's screenshot ("Start PV Power(kW)" inside the Smart Load tab), not from the wire name.
- **No 255 sentinel is assumed here.** AC Couple's END SOC has one; both our units read a plain `60` for Smart Load, so none was invented.


btli:
## Adversarial review round (Codex gpt-5.6-sol, xhigh)

**No P1 blockers.** The review's job here was mainly to attack the risky half of this PR — generalizing the shipped, three-rounds-reviewed `ac_couple_soc` store machinery instead of copying it. It compared the pre- and post-refactor AC-couple paths line by line and reported **CLEAN** on: the all-`None` predicate, the per-field carry-forward (including the `enabled` boolean type guard), the seed apply/clear/pop logic and per-field timestamps, the timeout, and the `ac_couple_{serial}` throttle key. The only difference it found is that the generic accessor creates an empty seed registry on first read, which changes no state behavior.

Also **CLEAN**: store isolation in both directions, the absent-value contract (missing / `None` / non-numeric / bool-as-int all stay unavailable; `0` stays a real value), write correctness, and entity lifecycle.

### Three findings fixed (`8d4b3e1`), each mutation-verified

1. **The spec table's `store_key → cloud_method` mapping was not pinned** — only the uniqueness of each column, and only three of the five specs had a write test. Swapping the End SOC and End Voltage writers left the suite **green** while making End SOC writes change End Voltage; since both take overlapping 0-100 inputs, it would have looked like it worked. Now the exact mapping is asserted and a parameterized write-routing test covers all five. *Mutation: the swap fails 3 tests.*
2. **`test_write_seed_survives_an_inflight_cycles_snapshot` was vacuous** — it seeded `self.data` before carrying forward, so no stale snapshot ever existed and the seed overlay was not needed to pass. It now reproduces the race in order: snapshot → write → publish the stale snapshot → carry forward. *Mutation: deleting the overlay fails it, and the AC-couple equivalent.*
3. **The optimistic-float test fed an already-rounded `53.5`**, so it passed with the rounding deleted. Now `53.56 → 53.6`. *Mutation-verified.*

### One finding deliberately declined

`_coerce_int_in_range`'s **0.01 integer tolerance** means `69.009` writes `69` rather than raising, which the review read as contradicting the "fractional values are rejected" contract. Tightening it to `is_integer()` would regress exactly what the tolerance exists for — float noise from templates and HA's own float transport, where `69.000000001` is indistinguishable to a user from a whole number — and the helper is **shared with every other percent control**, so the change would not be local to #499. The boundary is now pinned by a test instead, making it documented behavior rather than an accident.

### Also fixed from self-review (`e7aa1c5`)

`native_value` returned a `float` for the optimistic value but an `int` for the settled store value, so the SOC pair's state flipped `"80.0"` → `"80"` the moment the store caught up — enough to break an exact-string automation condition. Both paths now share one shaping step.

**2406 tests pass**, ruff clean, mypy clean, all CI checks green.

btli:
## Externals round applied (`4241642`)

### 1. Throttle stamp-before-fetch — the real bug

`_fetch_cloud_param_store` stamped `_last_status_fetch` **before** awaiting the getter and never took it back, so **one failed read cost the full 5-minute interval**. Worst case is the first fetch after startup: nothing to carry forward, so every entity of that store sits unavailable for five minutes because of a single network blip.

The stamp is still taken up front — it is re-entrancy protection, stopping a second call in the same cycle launching a duplicate read — but the failure path now **rewinds** it so the next attempt is due after `CLOUD_PARAM_STORE_RETRY_FLOOR` (120 s) rather than immediately. Bounded rather than eager on purpose: a persistent cloud outage then retries at that spacing instead of on every ~20-30 s poll. Same trade and same value as the parameter refresh's `_PARAMETER_RETRY_INTERVAL` (#282).

An **all-`None` read is deliberately not a failure** — it returns with the full stamp, because the read did happen and a device that genuinely lacks the params answers that way every time; re-arming on it would poll them forever.

Fixed once in the **shared** method, so it covers `ac_couple_soc` as well as `smart_load`. Tested on **both** — only a test on the AC-couple store proves the shipped path got it too. *Mutation-verified: removing the rewind fails both tests.*

### 2. CHANGELOG / pin coherence

The changelog claimed 0.9.39b6 "is installed automatically" — false while the manifest pins `>=0.9.39b5`, which is already released and satisfies the constraint, so HA would never upgrade and an eg4-only upgrader would get seven permanently unavailable entities. Reworded: the release carrying this feature is what pins b6, and installing the integration alone is not enough. **The pin is still not bumped here** — release-cut owns it.

The suite also passed green against pylxpweb **0.9.36b24**, because every other test mocks the control endpoint. Added `test_smart_load_methods_exist_in_installed_pylxpweb`, which resolves the seven methods against the **installed** package, plus a companion test anchoring that list to the real call sites so a rename cannot leave the pin passing against methods nothing uses.

**That test is DELIBERATELY RED until b6 ships** — CI now reports exactly one failure with the intended message, alongside 2430 passing. Red means "the dependency is not satisfied yet", not "the integration is broken".

### 3. Wording

The changelog asserted the firmware "honours whichever matches its configured mode" and described the write path in the present indicative. **The recorded evidence is a read-only probe — nothing was written to a device.** The mode behaviour is now stated as what the reporter's screenshot suggests and explicitly untested; the write paragraph says plainly that writing has only been reasoned about, never observed, with the #484-style ask to validate. The same correction was applied to the `const/limits.py` and `SmartLoadNumber` comments that carried it.

**2430 pass, 1 deliberate red, ruff and mypy clean; every other CI check green.**

PR #520, opened 2026-08-02
fix: unwind failed entry setup

btli:
## Summary

- roll back partially forwarded platforms when config-entry setup fails
- shut down the coordinator and close its HTTP client after refresh, setup, or cancellation failures
- clear failed runtime data so later retries cannot reuse a dead coordinator
- track the zero-read LOCAL follow-up refresh as coordinator-owned background work

## Root cause

The coordinator acquired a Home Assistant stop listener, transports, background state, and (for cloud modes) a client before the first refresh. `runtime_data` was assigned only after that refresh, and setup had no general failure owner. An initial-refresh error or partial platform-forward failure therefore skipped the normal unload path. Separately, the LOCAL static phase created an untracked refresh task that could outlive setup and race transport shutdown.

## Verification

- setup/coordinator regression suite: 590 passed
- full test suite: 2547 passed, 3 skipped
- `ruff check .`
- `ruff format --check .`
- configured strict mypy: 41 source files clean
- pre-commit hooks

Tracks `eg4-06er.1` (audit INT-05 / INT-07).

### DIFF
--- a/custom_components/eg4_web_monitor/__init__.py
+++ b/custom_components/eg4_web_monitor/__init__.py
@@ -593,9 +593,62 @@ def _async_cleanup_stale_smart_port_entities(
     return pending_serials
 
 
+async def _async_cleanup_failed_entry_setup(
+    hass: HomeAssistant,
+    entry: EG4ConfigEntry,
+    coordinator: EG4DataUpdateCoordinator,
+) -> None:
+    """Roll back resources acquired by an entry that did not finish setup."""
+    if coordinator._platform_setup_started:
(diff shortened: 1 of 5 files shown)


btli:
Follow-up pushed in 7d17595: cleanup now uses nested try/finally so a second CancelledError during shutdown or client close cannot strand later rollback steps or leave runtime_data attached. Added a regression covering cancellation during async_shutdown. Verification: 591 surrounding tests passed; mypy passed (41 files); pre-commit hooks passed.

PR #522, opened 2026-08-02
Expose safe read-only runtime diagnostics

btli:
## Summary

- expose safe, read-only LOCAL/HYBRID diagnostics from canonical input data: phase-aware EPS apparent power (I25), runtime (I69-I70), AC source (I77 bit 0), and parallel topology (I113)
- gate I25 on positively resolved phase context: phase-neutral on known non-three-phase systems, explicitly R-phase on known three-phase systems, and hidden when topology is unresolved
- keep I69-I70 as a disabled diagnostic duration measurement; do not claim total-increasing statistics without reset/lifetime hardware evidence
- decode proven enum fields while preserving missing and standalone values as unknown
- keep every new entity disabled by default and diagnostic-only; no writable register surface is added

## Hardware-evidence decisions

- pinned ant0nkr commit `d3d1014` identifies I25 as off-grid apparent power and explicitly qualifies it as R-phase on three-phase systems
- pinned pylxpweb `v0.9.39b6` provides split-phase aggregate semantics for its canonical I25 field, including the I131 + I132 fallback when the legacy combined value is zero
- the integration therefore publishes `eps_apparent_power` only in known non-three-phase context, `eps_apparent_power_r` only in known three-phase context, and neither for unresolved family/topology; explicit user grid overrides carry the resolved topology into mapping and entity gating
- I69-I70 remains raw seconds, corroborated by both register tables; neither source nor available captures establish reboot, replacement, or reset continuity, so `measurement` is used instead of `total_increasing`
- only proven I77 bit 0 is mapped (`Grid` / `Generator`); comparison-only AC-couple bits 1-2 remain intentionally unmapped pending EG4 captures
- I113 uses pylxpweb hardware-tested zero-based phase decode (`0=R, 1=S, 2=T`), not the comparison implementation conflicting one-based labels
- I25 remains LOCAL/HYBRID because released pylxpweb has no public phase-contextual cloud property; the integration does not reach through private runtime state

## Regression coverage

- mapping exclusivity for known non-three-phase, known three-phase, and unresolved topology
- family/entity gating, including UNKNOWN-family explicit grid overrides
- LOCAL static-key union and register-contract provenance for both mutually exclusive I25 keys
- HYBRID split-phase, three-phase, and unresolved integration paths
- I69-I70 state-class policy, translations, diagnostic defaults, and writable-surface isolation

## Verification

- `pytest -q tests`: 2558 passed, 3 skipped
- focused coordinator/diagnostic suite: 395 passed
- `ruff check .`: passed
- `ruff format --check .`: 117 files already formatted
- strict mypy: no issues in 41 source files
- `prek run --all-files`: all hooks passed

## Tracking

- bead: `eg4-uwa0.2`
- evidence-policy correction: `eb78a46`
- companion audit: register/race/performance gap report in a separate documentation PR


btli:
Resolved the automated review localization gap in follow-up 1740e2b: all six runtime diagnostics now use translation keys, enum states are stable lowercase slugs, strings/state maps are present with parity across all 13 locale files, and mapping/docs/tests were updated. Verification: 2,549 passed, 3 skipped; Ruff/format, strict mypy, JSON/locale parity, and pre-commit all pass.

btli:
Final evidence-policy verification for `eb78a46`:

RCA:
- I25 was previously exposed under a universal phase-neutral/aggregate label. Pinned ant0nkr `d3d1014` explicitly says I25 is R-phase on three-phase systems, while pinned pylxpweb `v0.9.39b6` supplies split-phase aggregate behavior. The correction selects `eps_apparent_power` only for positively known non-three-phase context, selects `eps_apparent_power_r` only for positively known three-phase context, and publishes neither when topology is unresolved. Explicit grid overrides now preserve the topology used by mapping and entity gating.
- I69-I70 had `total_increasing`, but the available evidence establishes only a 32-bit seconds unit. No source or hardware capture establishes reset, reboot, replacement, or lifetime continuity. It remains an opt-in diagnostic duration with `measurement`; promotion to a statistics total is deferred pending real capture evidence.

Safety and propagation:
- phase context reaches both LOCAL runtime mapping and the HYBRID transport overlay
- entity creation independently fails closed for ambiguous I25 context
- mapping, family/override, entity, HYBRID, static-key, register-contract, translation, and state-class regressions cover the policy
- no writable mapping or control surface changed
- independent production-diff review found no blocker

Verification:
- full local suite: 2558 passed, 3 skipped
- focused coordinator/diagnostic suite: 395 passed
- Ruff check and format: passed
- strict mypy: no issues in 41 source files
- prek all-files: all hooks passed
- replacement GitHub CI: HACS/Hassfest plus Bronze, Silver, Gold, and Platinum tiers all passed

The PR remains draft and was not merged.

PR #524, opened 2026-08-02
Document register, race, and performance audit

btli:
## Summary

This documentation-only PR is the complete EG4 integration / `pylxpweb` audit and remediation handoff requested in `eg4-vy1b`.

- records **29 scan lanes** (26 parallel external scans plus three reconciliation/validation lanes) across lifecycle, races, transport writes, cloud auth/session ownership, mappings, entities, performance, and CI
- provides evidence-backed RCA, impact, confidence, affected boundaries, rejected claims, and deterministic probes for every retained finding
- compares mapping/register behavior against pinned `ant0nkr/luxpower-ha-integration` commit `d3d101498bc...` and the exact released `pylxpweb` floor `ee16a6a`
- separates implemented safe read-only additions from disputed, family-gated, or hardware-capture-blocked registers/bits
- indexes **18 isolated integration remediation drafts** (#517-#535 excluding this audit PR) and **four dependency drafts** (#260-#263), with verification evidence and merge boundaries
- documents the final high-risk composed rehearsal against pylxpweb #261: **2,630 passed, 3 skipped**, plus focused teardown/auth/listener/breaker/generation/session traces and strict quality gates

## Important scope

This branch changes documentation only. Production fixes live in their own draft PRs, all based on canonical `origin/main` `90c9807`; none were merged during this work. Speculative writable mappings remain unimplemented. Dependency fixes still require release and manifest adoption (`eg4-06er.14`).

## Mapping conclusion

More data can be mapped safely, but not by wholesale copying the reference integration. Implemented read-only additions are I67, phase-contextual I25, I69-I70, captured I77 bit 0, and I113 diagnostics. I10, I68, I77 bits 1-2, battery-slot limits/cutoffs, and all proposed writable H110/H233 additions remain blocked on semantics, family evidence, sentinel behavior, or write/restore captures.

## Review guide

Start with the executive summary, finding/remediation indexes, integration and merge boundaries, register-map comparison, and final assessment. The detailed RCA sections retain source pins and exclusions so unsafe assumptions are not reintroduced later.

## Tracking

- audit bead: `eg4-vy1b` (closed)
- dependency adoption gate: `eg4-06er.14` (open)
- evidence/UX follow-ups: `eg4-uwa0`, `eg4-06er.12`, `eg4-06er.13`

This PR remains draft and unmerged.

PR #528, opened 2026-08-02
Make local quality runners fail honestly

btli:
## RCA

The audit found four independent false-green or unusable local-quality paths:

- tests/pytest.ini declared rootdir as an ini key even though pytest only supports it as a CLI option, producing a warning on every suite run.
- tests/run_tests.py searched for requirements-test.txt at the repository root even though the maintained file lives beside the runner in tests/.
- The runner probed optional Flake8 and returned success when it was absent, while project CI and pre-commit use Ruff as the canonical linter/formatter.
- The Platinum validator treated mypy errors or an unavailable mypy executable as success as long as strict=True existed in the config.
- Maintained development guidance named a nonexistent Bronze validator, an incorrect mypy config path, uv sync despite no project metadata, and coverage targets that differed from CI.

Signal-killed child processes also returned negative codes that main could collapse to success through max(0, returncode).

## Fix

- Remove the unsupported pytest ini key and make the runner pass its config explicitly.
- Resolve tests/requirements-test.txt beside the runner and fail when it is missing.
- Run Ruff check plus Ruff format --check through the current Python environment; propagate every nonzero or execution failure.
- Normalize signal termination to a shell-safe failing exit code.
- Run strict mypy through the validator Python environment and fail validation on errors, absence, or execution failure.
- Align local setup, test, coverage, typing, and tier-validation documentation with files and commands that exist in this repository.

## Regression coverage

A new quality-runner suite proves:

- the exact requirements path;
- exact Ruff lint and format commands;
- lint, format, unavailable-tool, and signal exit propagation;
- mypy error/unavailable/success outcomes;
- supported pytest ini keys;
- every documented validator path exists.

The original RED run failed 8 of 9 tests for the audited defects.

## Verification

- Runner-driven full suite: 2,553 passed, 3 skipped
- Quality runner regressions: 10 passed
- Real Ruff lint and format checks: pass
- Real Platinum validator including strict mypy: pass
- Independent strict mypy: 41 files clean
- Pre-commit hooks: pass
- Pytest configuration warning: absent
- Diff check: pass

Tracking: eg4-06er.8
Audit/RCA: PR #524

### DIFF
--- a/.github/WORKFLOWS.md
+++ b/.github/WORKFLOWS.md
@@ -103,8 +103,8 @@ Claude Code Review (1 job, PR only)
 You can run individual validation scripts locally:
 
 ```bash
-# Bronze tier
-python tests/validate_bronze_tier.py
+# Bronze tier checks run as dedicated jobs in quality-validation.yml;
+# there is no standalone Bronze validator.
 
 # Silver tier
 python tests/validate_silver_tier.py
@@ -112,8 +112,11 @@ python tests/validate_silver_tier.py
 # Gold tier
 python tests/validate_gold_tier.py
 
+# Platinum tier (includes an actual strict-mypy run)
+python tests/validate_platinum_tier.py
+
 # All tests with coverage
-pytest tests/ --cov=. --cov-report=term-missing
+pytest -c tests/pytest.ini tests/ --cov=custom_components/eg4_web_monitor --cov-report=term-missing
 ```
 
 ### Code Quality Checks
--- a/.github/WORKFLOWS.md
+++ b/.github/WORKFLOWS.md
@@ -116,7 +116,7 @@ python tests/validate_gold_tier.py
 python tests/validate_platinum_tier.py
 
 # All tests with coverage
-pytest -c tests/pytest.ini tests/ --cov=custom_components/eg4_web_monitor --cov-report=term-missing
+pytest tests/ --cov=custom_components/eg4_web_monitor --cov-report=term-missing
 ```
 
 ### Code Quality Checks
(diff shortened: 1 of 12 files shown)


btli:
CI follow-up pushed in c941084. RCA: the standalone Platinum validator now correctly propagates python -m mypy failures, but its job used bare Python 3.12 and installed no dependencies; the separate strict-typing job passed because it installed mypy and the import graph. The job now uses Python 3.13 and installs tests/requirements-test.txt, which now declares mypy, and a regression test locks that workflow/dependency contract. Verification: focused runner tests 11 passed; full suite 2554 passed, 3 skipped; strict mypy passed for 41 source files; all hooks passed; and a clean Python 3.13 environment installed the exact requirements and passed tests/validate_platinum_tier.py.

btli:
Additional runner-contract follow-up published in 7bb86a3. RCA 1: bare repository-root pytest used default collection because the only pytest.ini lived under tests/, so scripts/collision_test.py was imported and parsed pytest CLI flags at module import. RCA 2: once that import was made safe, root pytest still missed asyncio_mode=auto, yielding 2,558 async-fixture errors. The canonical config now lives at repository root with testpaths=tests; the helper, workflows, disabled workflow reference, and maintained docs all use it. The hardware collision script parses its rounds argument only under __main__. Added exact helper-command/config/import regressions and removed a dead exit-code fallback. Verification: bare root pytest -q is green with 2,556 passed / 3 skipped; helper path 13 passed / 2,546 deselected; Ruff, strict mypy (41 files), prek, and diff checks pass.

PR #529, opened 2026-08-02
fix: serialize shared physical endpoint operations

btli:
## Summary

- own one task-reentrant operation lock per physical network endpoint or serial adapter
- share that lock across logical devices and separate EG4 config-entry coordinators in the same Home Assistant instance
- bind HYBRID/LOCAL/legacy polls, direct writes, deferred parameter reads, quick-charge reads, parameter refreshes, and buffer drains to the shared endpoint boundary
- keep independent endpoints concurrent and preserve concrete pylxpweb transport identity

## Root cause

The coordinator already grouped normal polls by endpoint, but that only serialized poll-versus-poll work inside one update coroutine. pylxpweb correctly protects multi-register operations with a task-reentrant `_op_lock`, but each logical device receives a separate transport object and therefore a separate lock.

That left physical single-slot hardware unprotected across object boundaries: a poll for device A could overlap a named/raw write, deferred parameter load, quick-charge register read, parameter refresh, or startup buffer drain for device B. Separate config entries could also construct independent coordinators for the same endpoint and bypass a coordinator-local registry.

## Design

`EndpointOperationLock` mirrors pylxpweb's task-reentrant operation-lock contract. The integration installs one Home Assistant-scoped instance into every pinned pylxpweb transport sharing a normalized `host:port` or serial-port key. Reusing pylxpweb's existing operation seam has three useful properties:

- complete transport operations remain atomic, including nested named-parameter read/modify/write calls
- concrete `ModbusTransport`, `ModbusSerialTransport`, and `DongleTransport` identity is preserved; no proxy changes `isinstance`, attributes, mocks, or shutdown de-duplication
- steady-state work pays only the existing transport lock acquisition; repeated binding is an identity check, and different endpoints still run concurrently

Reconnect plus write is held as one endpoint transaction so two disconnected device objects cannot dial a single-slot endpoint together. Disconnect is deliberately not held behind the operation lock: coordinator shutdown must close the socket immediately to interrupt a hung read before cancelling background tasks.

## Regression coverage

The deterministic fake single-slot endpoint covers:

- poll versus write on one TCP endpoint
- two devices on one serial adapter
- deferred parameter work versus write
- reconnect plus write on disconnected device transports
- separate config-entry coordinators sharing one physical endpoint
- the deprecated single-device transport path
- shutdown interrupting a lock-held read
- task-reentrant nested operations
- concrete transport behavior after repeated configuration
- continued concurrency for independent endpoints

## Verification

- `pytest -q tests` — **2556 passed, 3 skipped**
- endpoint/group focus — **14 passed**
- coordinator/local/write-path focus — **592 passed** before the final added reconnect regression; the full suite above includes it
- `mypy --config-file tests/mypy.ini custom_components/eg4_web_monitor/` — success, 42 source files
- `prek run --all-files` with the project virtualenv on `PATH` — all hooks passed (YAML, EOF, whitespace, large files, Ruff, Ruff format, mypy)
- `git diff --check` — clean

## Companion PR context

- Audit/RCA index: #524 (ARC-01)
- Mixed transport cadence/starvation fix: #523; orthogonal to this physical-operation boundary
- pylxpweb atomic Quick Charge/H233 logical transaction: joyfulhouse/pylxpweb#260; complementary per-device RMW protection, while this PR protects across transport instances sharing hardware

No register mappings, entity behavior, or polling intervals change in this PR.

## Terminal lifecycle reconciliation

A paired integration/dependency review found that the coordinator only selected transports whose `is_connected` flag was already true. A dongle still inside `connect()` deliberately keeps that flag false until the socket and initial-data phase are usable, so entry unload could skip the terminal seam and allow the old transport to finish connecting after reload. The station de-duplication pass also pre-marked cached identities before proving shutdown had run.

The final shutdown pass now:

- collects legacy, inverter-cache, MID-cache, and station-attached transports before dispatch
- de-duplicates object identities across all four sources
- invokes a concrete `async_shutdown()` regardless of the transient connected flag
- retains reusable `disconnect()` only for older transports that report a live connection

Deterministic regressions cover cached and uncached station transports with a connect held in flight, cache/station identity duplication, no resurrection, and clients without the optional terminal method.

Companion dependency PR joyfulhouse/pylxpweb#263 makes the terminal flag authoritative across connect and request retry boundaries.

Reconciliation verification at `d71c680`:

- shutdown/endpoint focus: **24 passed**
- full integration suite: **2556 passed, 3 skipped**
- Ruff and Ruff format: passed
- strict mypy: **42 source files clean**
- all prek hooks: passed
- `git diff --check`: clean

### DIFF
--- a/custom_components/eg4_web_monitor/coordinator.py
+++ b/custom_components/eg4_web_monitor/coordinator.py
@@ -109,9 +109,14 @@ from .coordinator_mixins import (
     is_transport_link_down as _device_transport_link_down,
 )
 from .const.sensors import SENSOR_TYPES
+from .transport_serialization import (
+    EndpointOperationLock,
+    physical_endpoint_key,
+)
 from .utils import async_write_with_cloud_fallback
 
 _LOGGER = logging.getLogger(__name__)
+_ENDPOINT_OPERATION_LOCKS_DATA = f"{DOMAIN}_endpoint_operation_locks"
 
 PV_STRING_LIFETIME_STORAGE_VERSION = 1
 PV_STRING_LIFETIME_STORAGE_KEY = f"{DOMAIN}_pv_string_lifetime"
@@ -274,6 +279,13 @@ class EG4DataUpdateCoordinator(
             CONF_LOCAL_TRANSPORTS, []
         )
         self._local_transports_attached = False
+        # One lock per physical host:port or serial adapter. pylxpweb's
+        # operation lock is intentionally per transport instance, while this
+        # Home Assistant-scoped registry owns the physical boundary across
+        # logical devices and separate EG4 config-entry coordinators alike.
+        self._endpoint_operation_locks: dict[str, EndpointOperationLock] = (
+            hass.data.setdefault(_ENDPOINT_OPERATION_LOCKS_DATA, {})
+        )
         # Serials whose local-transport attach failed (commonly: the dongle's
         # single TCP slot still held by the previous session right after an
         # HA restart). Retried with a bounded interval each update cycle and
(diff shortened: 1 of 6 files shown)


btli:
Cross-repository shutdown-contract follow-up published in e108fdf, paired with joyfulhouse/pylxpweb#263 commit 0e746ff. RCA: pylxpweb correctly serialized reusable DongleTransport.disconnect with in-flight transactions, but coordinator shutdown calls disconnect before cancelling the task that may own that lock, creating an unload deadlock until the network timeout. The coordinator now feature-detects the concrete terminal async_shutdown seam, which closes the socket without the held operation lock and prevents reconnect; older/ordinary transports retain disconnect fallback. A deterministic fake makes reusable disconnect block behind the held endpoint lock and proves terminal shutdown is selected instead. Verification: 2,553 passed / 3 skipped; 21 focused shutdown/endpoint contracts; strict mypy (42 files), Ruff, prek, and diff checks pass.

btli:
Terminal lifecycle reconciliation is pushed at `d71c680`.

The coordinator now discovers all legacy/cache/station transport objects before one identity-de-duplicated shutdown pass. Concrete terminal-capable transports receive `async_shutdown()` even while `is_connected` is false during a pending dial; older clients retain connected-only reusable `disconnect()`.

New deterministic coverage holds both cached and uncached station connects in flight, proves cache/station duplicates shut down exactly once without resurrection, and covers the no-`async_shutdown` fallback.

Fresh verification: shutdown/endpoint focus 24 passed; full suite 2556 passed, 3 skipped; Ruff/format; strict mypy 42 files; all prek hooks; diff check. Companion retry-boundary fix: joyfulhouse/pylxpweb#263. Left as draft and unmerged.

PR #530, opened 2026-08-02
fix: prevent duplicate cloud plant coordinators across modes

btli:
## RCA

HTTP entries used `username_plant`, while HYBRID entries used `hybrid_username_plant`. Home Assistant deduplicates config flows by exact config-entry unique ID, so the same remote account+plant could be configured once in each mode. Both coordinators then polled the same station and competed for the same entity/device identities.

Reconfigure updated connection data and mode but never updated the config-entry unique ID, leaving identity stale after account and mode changes.

## Identity and migration policy

- Canonical cloud identity is the existing HTTP form, `username_plant`, for both HTTP and HYBRID. This minimizes churn for existing HTTP installs.
- Config creation and reconfigure compare both exact unique IDs and canonical identity derived from stored entry data, so legacy `hybrid_` IDs cannot bypass ownership.
- Reconfigure performs its final conflict scan immediately before the synchronous registry write. Conflicts abort without changing data, options, title, or unique ID.
- Config flow version is bumped to 3. Migration stages all changes before committing.
- An existing exact canonical owner wins. If no canonical ID exists, the oldest matching entry wins with entry ID as deterministic tie-breaker.
- Losing duplicates return migration failure with version, data, and unique ID untouched. No entry is deleted, merged, or overwritten.
- Existing local-only IDs are preserved because historic local display names were not persisted. Only a real cloud-to-local transition derives a new local ID and releases the cloud identity.
- The elected canonical owner remains reconfigurable while a rejected data-only legacy duplicate remains in the registry.

This selection is independent of setup order and prevents two migrations from electing different owners. Migration and registry updates contain no awaits between ownership decision and mutation on the Home Assistant event loop.

## Tests

- Cross-mode onboarding rejection, including legacy mode-prefixed IDs
- Canonical HTTP/HYBRID creation
- HTTP to HYBRID, HYBRID to HTTP, local to HYBRID, and HYBRID to local identity behavior
- Cloud account identity updates
- Non-destructive exact and legacy-data conflict rejection
- Canonical-owner usability with a rejected duplicate
- V1 to V3 atomic migration
- V2 hybrid canonicalization and integer plant normalization
- Existing canonical-owner election and reverse-order duplicate migration
- Oldest-owner election when no canonical ID exists
- Existing local identity preservation

## Verification

- `pytest -q tests`: 2553 passed, 3 skipped
- Focused config/migration suite: 173 passed
- Ruff check and format: passed
- Strict mypy: passed for 41 source files
- Changed-file hooks: passed
- `git diff --check`: passed

The suite reports only pre-existing pytest configuration, deprecation, and AsyncMock warnings tracked by separate audit follow-up work.

## Boundaries

Canonical matching intentionally preserves existing semantics: username and plant ID are exact strings, and base URL is not part of identity. Duplicate losers are preserved for manual recovery; removing the elected owner allows a preserved loser to migrate on a later retry.

Bead: eg4-06er.7


btli:
Addressed the review request in bc8d783: added the Unreleased changelog entry documenting the duplicate-coordinator RCA and non-destructive migration behavior. I also made the no-await migration-election invariant explicit beside the staged update. Focused 173-test suite, strict mypy, Ruff/format, changed-file hooks, and diff checks remain green.

PR #531, opened 2026-08-02
Fix multi-entry logging and registry cleanup lifecycle

btli:
## RCA

The pylxpweb logger is process-global, but each config-entry setup wrote its level directly. Load order therefore decided the effective preference, failed setup leaked the chosen level, and unload never reconciled the remaining entries.

Reconfigure reloads also retained registry records for serials and plants removed from configuration. Blind cleanup would be unsafe because local startup battery data is incomplete and a physical device can be shared by multiple entries.

## Changes

- Track per-entry library-debug ownership in Home Assistant scope; DEBUG wins if any loaded entry requests it, quiet entries compose at WARNING, failed setup rolls back, failed unload retains ownership, and the final unload restores the pre-integration level.
- After a successful first refresh, compare physical serial/station roots against current processed data, configured local transports, and authoritative cloud station membership.
- Prune only definitively absent roots and their descendant device tree; never infer missing batteries from the local static startup snapshot.
- Remove only this entries entities/config ownership from shared registry devices.
- On config-entry deletion, deterministically remove owned entity/device records after statistics purge while preserving identities owned by another entry.

## Verification

- 7 focused lifecycle regressions passed
- Full suite: 2550 passed, 3 skipped
- Ruff lint and format passed
- Strict mypy passed for all 41 integration source files
- Full prek suite passed
- git diff --check passed

Audit: #524
Tracking: eg4-06er.10

### DIFF
--- a/custom_components/eg4_web_monitor/__init__.py
+++ b/custom_components/eg4_web_monitor/__init__.py
@@ -1,6 +1,7 @@
 """EG4 Web Monitor integration for Home Assistant."""
 
 import asyncio
+from dataclasses import dataclass, field
 import logging
 from typing import Any, TypeAlias
 
@@ -24,6 +25,8 @@ from homeassistant.helpers.storage import Store
 from .const import (
     CONF_CONNECTION_TYPE,
     CONF_HTTP_POLLING_INTERVAL,
+    CONF_LIBRARY_DEBUG,
+    CONF_LOCAL_TRANSPORTS,
     CONF_SENSOR_UPDATE_INTERVAL,
     CONNECTION_TYPE_HTTP,
     DEFAULT_HTTP_POLLING_INTERVAL,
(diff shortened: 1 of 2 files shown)


btli:
Review follow-up:\n\n- Tidied the test import grouping in daf4678; focused test_init (32 tests), Ruff, format, mypy hook, and all commit hooks pass.\n- I deliberately did not delete a childless parallel-group root here. This cleanup runs against first-refresh data, and LOCAL startup may be incomplete. Guessing group absence would recreate the destructive identity behavior fixed by #517. I filed eg4-06er.13 to add group pruning only after deriving authoritative HTTP/LOCAL topology, with shared-entry and ambiguity coverage.\n- Cleanup is reached only after async_config_entry_first_refresh returns successfully. A raised Station/load/refresh failure therefore bypasses registry mutation; authoritative station membership is also included to protect a device omitted only by downstream processing.\n\nThat keeps this PR focused on provable station/serial absence while making the remaining group-root edge explicit and test-gated.

PR #532, opened 2026-08-02
perf: scope coordinator listener fanout

btli:
## RCA\n\nEvery coordinator refresh previously notified every registered listener, even when only one inverter changed or the refresh merely carried forward unchanged cached data. Late-discovery callbacks then rescanned the complete device collection. With hundreds of entities, routine polling therefore produced callback and discovery work proportional to the whole installation.\n\n## Fix\n\n- Route coordinator updates by station, discovery, and device listener contexts.\n- Preserve full fanout for initial data, availability transitions, unknown top-level changes, and deliberately unscoped listeners.\n- Detect carried-forward in-place mutations safely by snapshotting the previous coordinator data.\n- Limit late discovery to serials whose active device data changed.\n- Scope read entities and firmware updates to their source device.\n- Keep controls and optimistic entities unscoped so retention TTL and availability transitions cannot be skipped.\n- Scope manual parameter/cloud acknowledgements to the affected serial plus discovery.\n\n## Performance evidence\n\nA 300-listener regression test confirms that a FAST-only device change invokes the 150 FAST listeners while skipping all 150 SLOW listeners, while preserving discovery and unscoped safety listeners. A synthetic 10-device snapshot/compare benchmark with 500 sensors and 6 batteries per device measured about 1.568 ms for two deep copies plus comparison across 500 iterations; production performs one snapshot.\n\n## Verification\n\n- 2,550 passed, 3 skipped\n- focused coordinator coverage: 633 passed\n- focused platform coverage: 136 passed\n- strict mypy: 41 source files clean\n- Ruff lint and format checks pass\n- prek hooks pass\n- git diff --check passes\n\nBead: eg4-06er.11\n\nThis PR is intentionally isolated and remains a draft for review.

### DIFF
--- a/custom_components/eg4_web_monitor/base_entity.py
+++ b/custom_components/eg4_web_monitor/base_entity.py
@@ -36,7 +36,11 @@ from .const import (
     MANUFACTURER,
     SENSOR_TYPES,
 )
-from .coordinator import EG4DataUpdateCoordinator
+from .coordinator import (
+    STATION_LISTENER_CONTEXT,
+    EG4DataUpdateCoordinator,
+    device_listener_context,
+)
 from .utils import (
     async_write_with_cloud_fallback,
     clean_model_name,
@@ -80,7 +84,7 @@ class EG4DeviceEntity(CoordinatorEntity):
             coordinator: The data update coordinator.
             serial: The device serial number.
         """
-        super().__init__(coordinator)
+        super().__init__(coordinator, context=device_listener_context(serial))
         self.coordinator: EG4DataUpdateCoordinator = coordinator
         self._serial = serial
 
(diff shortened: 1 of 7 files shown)


btli:
Reviewer follow-up published in d8ba709. RCA: the original tuple scope was not private at runtime, so foreign hashable contexts could be silently filtered and foreign unhashable contexts could abort dispatch. The pending empty-set sentinel also leaked across public/direct dispatches, including while transport I/O was awaiting, starving integration-owned listeners outside a classified refresh. The fix uses a private runtime context type, treats all foreign contexts as default listeners, keeps None as the idle/full-fanout sentinel, and stages a scoped set only after a concrete result is classified. Added direct idle, in-flight I/O, foreign hashable/unhashable, repeated-failure, and unchanged-scope regressions. Verification: 2,553 passed / 3 skipped; focused 10 passed; Ruff, strict mypy (41 files), prek, and diff checks pass. Independent re-review found no remaining correctness blocker.

PR #533, opened 2026-08-02
fix: bound and deduplicate cloud polling

btli:
## Summary

- enforce one Home Assistant-local, exact-account cloud request budget across every config entry/plant for that account; nested pylxpweb startup, refresh, parameter, and supplemental fanout is capped at three request chains in aggregate
- make pylxpweb #261 reactive authentication re-entrant without deadlock or capacity escape by handing one admitted parent lease to the shared auth task until login, account detection, and retry/backoff finish
- replace broad post-write refreshes with serial-scoped parameter reads and one in-place listener publication; parallel-group refreshes publish all changed siblings in one batch
- single-flight missing-parameter jobs and account-wide firmware progress, with shutdown/reload-safe ownership transfer and bounded cleanup
- preserve the explicit Refresh button's full runtime/energy/battery/parameter behavior while removing duplicate number/select prefetches

This closes bead `eg4-06er.5` (PERF-01..03 and reconciliation findings).

## Root cause analysis

1. `_api_semaphore` guarded per-device mapping only after pylxpweb had entered `Station.load()` / `Station.refresh_all_data()`. Their nested gathers created runtime, energy, battery, parameter, and supplemental HTTP calls before the integration semaphore could see them, so concurrency scaled with inverter and entry count.
2. Parameter verification called broad `refresh(force=True, include_parameters=True)` or performed entity-level prefetch plus coordinator refresh. A holding-register write therefore re-read unrelated runtime/energy/battery data, refreshed sibling devices, and could dispatch listeners repeatedly.
3. Missing-parameter scheduling used callback-driven follow-up tasks without a terminal shutdown gate. A normally completed task's queued done callback could recreate work after shutdown had snapshotted the task set.
4. Firmware progress is account-wide, but pylxpweb exposes it per device/client. Aligned cache expiry caused one raw status call per inverter and separate entries for the same account duplicated it again. Unloading the originating client could also invalidate a flight still awaited by a sibling entry.
5. pylxpweb #261 creates a shielded authentication child task. A plain semaphore deadlocks when three admitted parents await that child; a bare login bypass can instead outlive canceled parents and transiently permit a fourth raw operation.

### Account request budget

- `CloudRequestLimiter` wraps each client's common `_request` boundary, while `SharedCloudRequestBudget` stores only the semaphore/lifecycle state in `hass.data`, keyed by normalized `(username, base_url, verify_ssl)`.
[...]

PR #535, opened 2026-08-02
fix: isolate cross-entry cloud cookies

btli:
## Summary

- give every coordinator/config-flow cloud client its own cookie jar while retaining Home Assistant's connector, resolver, SSL context, and user-agent
- pass `CONF_VERIFY_SSL` into Home Assistant's session factory so the injected session actually uses the selected SSL policy
- drain dependency-owned authentication work before detaching the private session on unload, Home Assistant stop, config-flow cancellation, and repeated shutdown cancellation
- retain Home Assistant's config-entry / `EVENT_HOMEASSISTANT_CLOSE` cleanup fallback for constructor failures
- update the Platinum validation checks to require the cookie-safe session factory

## Root cause

`async_get_clientsession(hass)` returns one process-wide session for each SSL/family tuple. aiohttp stores cookies by origin/domain/path, not by `LuxpowerClient`. Two EG4 accounts using the same portal origin therefore shared one `CookieJar`: account B's login replaced account A's session cookie while A's client still considered its independent two-hour expiry valid. Config-flow validation could overwrite a loaded entry in the same way.

The injected session also owns the effective connector. Passing `verify_ssl=False` only to `LuxpowerClient` had no effect because pylxpweb uses that flag when it creates a connector itself; the integration had already injected Home Assistant's default verified session.

Independent follow-up review found a second lifecycle defect: cancellation during `LuxpowerClient.__aenter__()` bypasses `__aexit__()`. pylxpweb #261 intentionally shields the client-owned authentication task, so the original config-flow `finally` could detach its session while authentication still used it. A second cancellation could similarly interrupt coordinator `client.close()` and jump directly to detach.

Home Assistant explicitly documents cookies as the case for [`async_create_clientsession`](https://developers.home-assistant.io/docs/core/integration-quality-scale/rules/inject-websession/). Its helper creates a distinct `ClientSession`/`CookieJar` but reuses Home Assistant's managed connector pool.

## Lifecycle contract

- coordinators use `async_create_clientsession(..., verify_ssl=..., auto_cleanup=True)` for a private jar plus config-entry / HA-close constructor-failure fallback
- normal unload and HA stop first run transport/background/request-budget/firmware cleanup, then call `LuxpowerClient.close()`, then detach only the session wrapper
- config-flow validation binds the client before explicit login and uses `auto_cleanup=False`
- one shared cleanup primitive waits for the exact client-close operation through repeated caller cancellations, observes its result/exception, detaches last, and propagates cancellation only after cleanup settles
- cancellation-neutral `asyncio.wait` avoids Python 3.14 event-loop leakage caused by repeatedly shielding a later-failing cleanup task
- detaching never closes Home Assistant's shared connector; repeated stop/unload cleanup is idempotent

## TDD and verification

The original cookie-isolation regressions failed before implementation. Two reviewer-follow-up cancellation regressions also failed on the first PR head and pass on `14b7488`:

- cancellation during config-flow login drains a shielded auth task before detach
- two caller cancellations plus a failing close preserve close-before-detach order, propagate cancellation with the cleanup error chained, and emit no event-loop error context

Final verification against pylxpweb PR #261 commit `61b7992`:

- `pytest -q tests`: **2,560 passed, 3 skipped**
- session isolation/lifecycle file: **17 passed**
- config/setup/coordinator compatibility selection: **597 passed**
- strict mypy: **42 source files clean**
- Ruff lint/format, `prek run --all-files`, and `git diff --check`: passed
- independent review: approved after **24 focused tests** plus adversarial close-error, child-cancellation, synchronous-close, repeated-cancellation, and event-loop leakage probes

## Composition boundaries

- **#520 setup unwind:** `coordinator.async_shutdown()` owns cloud auth/session cleanup. Its later defensive `client.close()` is idempotent; failed setup therefore drains before Home Assistant's session fallback callback.
- **#531 logging/registry lifecycle:** no state-key overlap. Its normal-unload `client.close()` is likewise a safe redundant close after coordinator shutdown.
- **#533 cloud fanout:** preserve the private session when reconciling the coordinator constructor. Its `super()` stop/shutdown path releases request-budget and firmware-flight ownership before client close/detach; constructor rollback retains Home Assistant's fallback.
- **pylxpweb #261 auth single-flight:** client-owned shielded authentication is cancelled/drained before detach. Focused and full integration suites load the exact PR source via `PYTHONPATH`.

Tracks bead `eg4-06er.15`. This PR is intentionally draft and unmerged.

### DIFF
--- a/.github/workflows/quality-validation.yml
+++ b/.github/workflows/quality-validation.yml
@@ -742,15 +742,15 @@ jobs:
         run: |
           echo "Checking websession injection support..."
 
-          # Check coordinator uses aiohttp session injection via async_get_clientsession
-          if ! grep -q "async_get_clientsession" custom_components/eg4_web_monitor/coordinator.py; then
+          # Cookie auth requires an isolated HA session per client/account.
+          if ! grep -q "async_create_clientsession" custom_components/eg4_web_monitor/coordinator.py; then
             echo "❌ Coordinator does not inject session"
             exit 1
           fi
           echo "✅ Coordinator injects Home Assistant session"
 
           # Check config flow uses session injection
-          if ! grep -q "async_get_clientsession" custom_components/eg4_web_monitor/_config_flow/__init__.py; then
+          if ! grep -q "async_create_clientsession" custom_components/eg4_web_monitor/_config_flow/__init__.py; then
             echo "❌ Config flow does not inject session"
             exit 1
           fi
@@ -764,7 +764,7 @@ jobs:
           echo "✅ LuxpowerClient API client is used"
 
           # Check that session is passed to LuxpowerClient
-          if ! grep -q "session=aiohttp_client" custom_components/eg4_web_monitor/coordinator.py; then
+          if ! grep -q "session=cloud_session" custom_components/eg4_web_monitor/coordinator.py; then
             echo "❌ Session not passed to LuxpowerClient"
             exit 1
           fi
--- a/custom_components/eg4_web_monitor/_config_flow/__init__.py
+++ b/custom_components/eg4_web_monitor/_config_flow/__init__.py
@@ -1390,29 +1390,44 @@ class EG4ConfigFlow(
 
     async def _test_cloud_credentials(self) -> None:
         """Test cloud credentials and load stations list."""
-        session = aiohttp_client.async_get_clientsession(self.hass)
         assert self._username is not None
         assert self._password is not None
-
-        async with LuxpowerClient(
-            username=self._username,
-            password=self._password,
-            base_url=self._base_url,
+        # Cookie-authenticated clients cannot use HA's process-wide default
+        # session: another account on the same origin would overwrite its
+        # domain cookie. The dedicated session still reuses HA's connector,
+        # resolver, SSL context, and user-agent. Config flows are not config
+        # entry setup, so explicitly detach instead of retaining the wrapper
+        # until Home Assistant stops.
+        session = aiohttp_client.async_create_clientsession(
+            self.hass,
             verify_ssl=self._verify_ssl,
-            session=session,
-        ) as client:
-            from pylxpweb.devices import Station
+            auto_cleanup=False,
+        )
+        try:
+            async with LuxpowerClient(
+                username=self._username,
+                password=self._password,
+                base_url=self._base_url,
+                verify_ssl=self._verify_ssl,
+                session=session,
+            ) as client:
+                from pylxpweb.devices import Station
 
-            stations = await Station.load_all(client)
-            # Normalize plant ids to str at the API boundary: pylxpweb returns
-            # Station.id as int, but the HA frontend submits form selections as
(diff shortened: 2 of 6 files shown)


btli:
Final verification for `e3f7c0e706016776e684017bfbaff51537cb14c2`:

- Product CI: **35/35 repository-owned checks passed**, including HACS, Hassfest, Gold coverage, Platinum comprehensive tests, strict typing, websession injection, and tier summaries.
- Local focused regression suite: **15 passed** (cookie isolation, SSL selection, config-flow isolation, constructor failure cleanup, config-entry fallback, `homeassistant_stop`, `homeassistant_close`, unload, ordering, and idempotency).
- Local full suite: **2558 passed, 3 skipped**.
- Local quality gates: Ruff lint/format, strict mypy across 41 source files, `prek run --all-files`, and `git diff --check` all passed.
- Compatibility run against pylxpweb PR #261 commit `61b7992`: **15 passed**.

The sole red check is the external Claude Code Review action. Its log reports `subtype: success`, `is_error: true`, one turn, `$0` cost, zero permission denials, and "No buffered inline comments"; it produced no PR comment or review. This is runner/provider infrastructure, not a repository test failure.

The PR remains draft and unmerged.

btli:
Cancellation-safety follow-up pushed as `14b748826a20f65dd600f765be2062796de2257e` after independent review found a real lifecycle gap in the original head.

RCA and resolution:

- Cancellation during `LuxpowerClient.__aenter__()` bypasses `__aexit__()`; with pylxpweb #261, the client-owned authentication task is shielded and can outlive that caller.
- The old config-flow `finally` could therefore detach its private session while authentication still used it. A second cancellation could likewise interrupt coordinator `client.close()` and jump to detach.
- Config flow now binds the client before login. Both config flow and coordinator use one cleanup primitive that waits for the exact close operation through repeated caller cancellations, observes its result/exception, detaches last, and propagates cancellation only after cleanup settles.
- The implementation uses cancellation-neutral `asyncio.wait`, avoiding Python 3.14's event-loop error logging from repeatedly shielding a later-failing task.

Final local evidence against the exact pylxpweb #261 source (`61b7992`):

- new deterministic regressions: **2 passed** (both failed on the previous head)
- session isolation/lifecycle file: **17 passed**
- config/setup/coordinator compatibility selection: **597 passed**
- full integration suite: **2,560 passed, 3 skipped**
- Ruff lint/format, strict mypy across **42 source files**, `prek run --all-files`, and `git diff --check`: passed
- independent reviewer: **approved**, including 24 focused tests and adversarial close-error, child-cancellation, synchronous-close, repeated-cancellation, and event-loop leakage probes

The PR remains draft and unmerged. Final-head GitHub CI is now running.

btli:
Final-head CI for `14b7488` is complete: **35/35 repository-owned checks passed**. The PR is `MERGEABLE` / `CLEAN`, remains draft, and has not been merged.

Issue #544, opened 2026-08-08
Generator power


brendonlobo123:
### Integration Version

Beta9

### Home Assistant Version

2026

### Connection Mode

Hybrid (Local + Cloud)

### Device Model(s)

12000XP

### Describe the Bug

After adding local connection it is showing generator power but I don't have anything connected 

sensor.sna_us_15k_60534a0404_generator_power

[config_entry-eg4_web_monitor-01KSWA215KHGP5HX9GJFN14SCF.json](https://github.com/user-attachments/files/30848241/config_entry-eg4_web_monitor-01KSWA215KHGP5HX9GJFN14SCF.json)

### Steps to Reproduce

Look at the entity readings. 

### Affected Entity ID(s)

sensor.sna_us_15k_60534a0404_generator_power

### Debug Logs (required)

[latest-4.txt](https://github.com/user-attachments/files/30848252/latest-4.txt)

### Diagnostics Download (required)

[config_entry-eg4_web_monitor-01KSWA215KHGP5HX9GJFN14SCF.json](https://github.com/user-attachments/files/30848245/config_entry-eg4_web_monitor-01KSWA215KHGP5HX9GJFN14SCF.json)

### Confirmation

- [x] I attached DEBUG-level logs captured while the problem was occurring (not just the default warnings log)
- [x] I attached the diagnostics download (or explained above why it is not possible)

brendonlobo123:
<img width="1080" height="2424" alt="Image" src="https://github.com/user-attachments/assets/22369a1b-4127-4006-9465-45dda53456fc" />

btli:
## Root cause found — and it's confirmed against your inverter's actual firmware

Short version: **your inverter's "Generator Power" register is not a power measurement at all.** On
the off-grid platform (12000XP / SNA-US, and the 6000XP) Modbus input register 123 returns an
internal **counter that ticks up once per second and rolls over at 65536**. It has nothing to do
with the GEN port. Your 28,646 W was the inverter telling us it had been running for 28,646
seconds — about 7 hours 57 minutes.

Your own two screenshots prove it, which is what made this click:

```
first reading   28,646     (diagnostics, Aug 7 20:31)
second reading   5,610     (screenshot,  Aug 8 08:19)

(5610 − 28646) mod 65536 = 42,500 seconds = 11h 48m 20s
actual gap between the two captures = 11h 48m 53s      → 33 seconds apart
```

[...]

PR #547, opened 2026-08-09
docs: EPS register firmware traces — off-grid and hybrid (#544 follow-up)

btli:
## Why

Two follow-up questions from #544, answered from the firmware rather than by inference:

1. After input register 123 turned out to be an ARM-local counter, **is EPS apparent power genuine?** (reporter's question)
2. **Does EG4_HYBRID carry the same registers?** (maintainer's question)

## Answers

**Off-grid: apparent power is genuine.** Inputs 25/131/132 resolve the DSP power block (`0x2000D0E4` +`0x88`/`0x8A`/`0x8C`) and trace end to end — C28x frame → parser (`0x08025AB8`) → filter (`0x08018D44`) → publish (`0x0801A2EC..0x0801A32C`) → FC04. The publish routine writes active and apparent power *alternating*, through the same filters, so the two carry identical confidence. The W÷10 / VA×1 asymmetry is confirmed by the firmware's own ×10 normalisation at `0x0800E3F4`.

**Hybrid: same register numbers, different meanings — and one is not a measurement.**

| Reg | Off-grid | Hybrid |
|---|---|---|
| 25 | direct DSP apparent power | ARM-computed `V × I ÷ 100`, clamped so S ≥ P |
| 131 | genuine apparent power | DSP power **split by sign into direction fields** |
| 132 | genuine apparent power | **an incrementing threshold-gated counter** |

Register 132 is the #544 pattern on a different family, and unlike the S/T voltages **it is exposed today** — `eps_apparent_power_l2` exists on both of the maintainer's hybrid inverters, reading `0 VA` only because EPS is idle and the counter's threshold hasn't been crossed. The doc suggests handling that mirrors the #544 fix (hybrid-scoped exclusion for 131/132, keep aggregate 25). **No code change in this PR** — that's a separate decision.

**EPS S/T voltage (21/22) is unusable in both families**, for different reasons. Off-grid: byte/bitfield composites. Hybrid: coherent DSP words that nothing uses as voltages, reading 256.0 V and 4832.0 V on two units. Currently suppressed everywhere by the three-phase gate — but hybrid is the three-phase-*capable* family, so that gate can open. Deliberately left in place: on real three-phase firmware those slots may carry real voltages, and no such hardware exists to test against.

## Method corrections (both fail silently)

- **`extract_firmware_from_pcap.py` output is still OTA-framed.** De-frame per *actual* chunk length (`455×771, 4×259, 481, 703, 7`), not a fixed period. Verified: 351,642 bytes, sha256 `63d0efba…`, and the 18kPV and FlexBOSS21 captures reconstruct **byte-identically**.
- **The hybrid ARM load base is `0x08010000`**, not the off-grid `0x08005000`. Derive it by back-solving a vector-table entry; confirm via RAM-pointer density and a known literal.

## On rigour

Every cited instruction and literal was verified byte-for-byte against the images before being recorded, and both traces were run past an independent adversarial pass.

Two of my own conclusions were **refuted and corrected in place** rather than quietly dropped:

- Off-grid: I first reported *zero writers* for the power fields and nearly concluded the struct was DMA-filled. Tooling artefact — the base pointer is held in `R6` across a ~6 kB function.
- Hybrid: I traced a byte-swapped report frame at `0x2000B6D4` that looks exactly like a register image and built a whole mechanism on it. It's a decoy; FC04 actually serves from `0x20009DBC` via the builder at `0x080594D8`.

Both traps are now written into `FIRMWARE_ACQUISITION.md`, because a wrong mechanism recorded as fact is how the `re/` artifacts became unusable in the first place.

## Scope

Docs only. No code, no behaviour change, no pin change.

- **new** `OFFGRID_EPS_REGISTERS.md`, `HYBRID_EPS_REGISTERS.md`
- `FIRMWARE_ACQUISITION.md` — working de-frame recipe, ARM load bases, scanner blind-spot warning
- `OFFGRID_GENERATOR_REGISTERS.md` — cross-links, scope banner

Answer posted to the reporter: https://github.com/joyfulhouse/eg4_web_monitor/issues/544#issuecomment-5229089082

### DIFF
--- a/docs/reference/firmware/FIRMWARE_ACQUISITION.md
+++ b/docs/reference/firmware/FIRMWARE_ACQUISITION.md
@@ -198,5 +198,15 @@ Anchor conclusions to measurements:
 - Where a register's meaning is at stake, check the decode against a live capture or an issue
   report before publishing it.
 
-See [`OFFGRID_GENERATOR_REGISTERS.md`](OFFGRID_GENERATOR_REGISTERS.md) for a full worked analysis
-following this method.
+Worked analyses following this method:
+
+- [`OFFGRID_GENERATOR_REGISTERS.md`](OFFGRID_GENERATOR_REGISTERS.md) — input 123 is an ARM-local
+  1 Hz counter, not generator power (issue #544).
+- [`OFFGRID_EPS_REGISTERS.md`](OFFGRID_EPS_REGISTERS.md) — inputs 25/131/132 (EPS apparent power)
+  *are* genuine DSP measurements, traced parser → filter → publish → Modbus; inputs 21/22 (EPS S/T
+  voltage) are byte/bitfield composites and not voltages.
+
+A scanner that reports "no writers" is the trap to watch for: a base pointer held in a callee-saved
+register across a large function defeats naive literal tracking, and the resulting silence reads
+exactly like proof of absence. Confirm the scanner can see a known-good writer before trusting it
+on an unknown one.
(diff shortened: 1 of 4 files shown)


brendonlobo123:
Grid always on disappeared. It's a switch in smart load



Book with me <https://calendar.app.google/sqymqbycKJdV8y1j6>

Issue #490, opened 2026-07-26
Internal temperature is 0


brendonlobo123:
### Integration Version

Latest stable

### Home Assistant Version

Latest stable

### Connection Mode

Cloud (HTTP)

### Device Model(s)

12000XP

### Describe the Bug

Internal temperature is 32 f

### Steps to Reproduce

Check the value of the inverter internal temperature 

### Home Assistant Logs

```shell

```

### Screenshots

<img width="1080" height="2424" alt="Image" src="https://github.com/user-attachments/assets/324ca783-ff25-440f-bf91-95c37fa26a76" />

<img width="1080" height="2424" alt="Image" src="https://github.com/user-attachments/assets/83da3e6f-7f8c-45e7-9c04-e6c8acfa8eee" />

PR #561, opened 2026-08-11
fix: blank constant-zero temperature registers on transport paths too (#560)

### LINKED ISSUE #560: Inverter shows freezing temperatures
### Integration Version

Beta9

### Home Assistant Version

2026

### Connection Mode

Hybrid (Local + Cloud)

### Device Model(s)

12000XP

### Describe the Bug

Freezing temperatures. I need to know where it is so I can get ice cubes in summer

Image

### Steps to Reproduce

Look at entity values

### Affected Entity ID(s)

_No response_

### Debug Logs (required)

[home-assistant_eg4_web_monitor_2026-08-09T12-52-43.860Z.log.txt](https://github.com/user-attachments/files/30873325/home-assistant_eg4_web_monitor_2026-08-09T12-52-43.860Z.log.txt)

### Diagnostics Download (required)

[config_entry-eg4_web_monitor-01KSWA215KHGP5HX9GJFN14SCF (1).json](https://github.com/user-attachments/files/30873326/config_entry-eg4_web_monitor-01KSWA215KHGP5HX9GJFN14SCF.1.json)

### Confirmation

- [x] I attached DEBUG-level logs captured while the problem was occurring (not just the default warnings log)
- [x] I attached the diagnostics download (or explained above why it is not possible)

### Screenshots

Image

btli:
## Root cause

A HYBRID-mode 12000XP (cloud model SNA-US 15K, deviceTypeCode 54, EG4_OFFGRID) serves a **constant 0** in input registers 64 (internal temp), 67 (battery temp) and 108 (BT temp) while radiator1/2 read a live 58/61 °C and BMS cell temps read a healthy 30/32 °C. Not a sentinel (127 → 0x7F already maps to None in pylxpweb), not signed-decode, not scaling — the DSP genuinely serves 0, so the three sensors rendered a permanent bogus 0 °C (32 °F, "freezing") on a running unit.

This is the missing evidence #490 lacked: the #490 fix (`blank_cloud_zero_internal_temperature`) exempted transport-backed values because we had no reg-64 reading from an off-grid unit. #560 shows the local register path serves the same constant 0 the cloud relays as `tinner: 0`, so the `has_transport_runtime` guard exempted exactly the reported case.

## Fix

Generalize the #490 blanking (`coordinator_mappings.py`, integration layer only — pylxpweb is correct):

- **Drop the transport exemption** — blanking now applies on CLOUD, HYBRID and LOCAL paths. The CLOUD/HYBRID call site moves after the transport overlay so an overlaid `bt_temperature` (reg 108) is treated too; the LOCAL path is covered inside `_build_runtime_sensor_mapping`.
- **Extend coverage** from `internal_temperature` to `battery_temperature` and `bt_temperature`.
- **Radiator corroboration**: an exact 0 is blanked to unknown only when a radiator reads a live nonzero value (a unit whose radiators read 58 °C is not at 0 °C ambient). Radiators also reading 0 are consistent with a genuinely cold unit and left alone; no radiator reading at all falls back to the plain value-scoped #490 behavior.
- **Still value-scoped, never family-gated**: the defect splits WITHIN deviceTypeCode 54 (#259/#307) — a 6000XP reports live `Tinner` of 31-32 °C while a 12000XP reports 0, both EG4_OFFGRID. Only the observed value is treated.

## Tests

Extended `tests/test_issue_490_internal_temperature.py`: unit coverage for the two new keys and all corroboration branches (live radiator / one live radiator / zero radiators / absent radiators / radiators never modified), LOCAL-path coverage via `_build_runtime_sensor_mapping` on a real `InverterRuntimeData`, and a #560 HYBRID end-to-end through `_process_inverter_object` (regs 64/67/108 at 0 + live radiators → unknown; live values survive).

Mutation-checked: stashing the src change turns the new tests red (collection error on the renamed symbol); behaviorally, neutering the blanking fails 8 tests (including both end-to-ends and the #490 regression) and removing the corroboration gate fails the 2 cold-climate tests; restored, all green.

## Risk note (cold-climate trade-off)

0 °C is a physically legitimate reading for an idle, unheated install. The corroboration gate bounds that false-positive: a genuinely cold unit reads ~0 at the radiators too and keeps its 0. The residual loss — a unit genuinely at 0 °C whose radiators nonetheless read live now reads "unknown" instead of 0 — is small, bounded, value-scoped only, and trivially reversible.

Gates: `ruff check`/`ruff format` clean, `mypy` clean (46 files), full suite 2832 passed / 3 skipped.

Fixes #560


btli:
## Tribunal round-1 fix wave (predicate redesign)

**Pushed head:** `500f5edb9d29e442dcc4f3554d0e1eefa9a77445`

### Reporter-log warmth confirmation (codex HIGH-3 gate)
Diagnostics from #560 show steady-state:
`internal_temperature=0`, `battery_temperature=0`, `bt_temperature=0.0`, **`radiator1_temperature=58`**, **`radiator2_temperature=61`**.
Live warm radiators → fix still applies to the reporter. Not BLOCKED.

### Core redesign
Blank constant-zero targets **only** when ≥1 radiator is **strictly `> 0` °C**. Radiators `≤0` (incl. negatives) and absent/`None` **protect** (publish the 0). All-zero boot frame publishes zeros as accepted residual (indistinguishable from genuine cold — no family/freshness heuristic).

### Per-finding
| Finding | Resolution |
|---|---|
| codex BLOCKER / HIGH-2 / claude MEDIUM-1 (predicate) | Redesigned to `> 0` warmth-only; absent no longer corroborates |
| codex HIGH-3 (all-zero residual) | Confirmed reporter warm; residual documented in docstring + DATA_MAPPING + test |
| claude test seam (negative radiators) | `test_negative_radiators_protect_zero`; `>0`→`!=0` mutation RED |
| claude LOW-2 (tri-state 0+None) | `test_tri_state_zero_and_none_radiators_publish` |
| codex MEDIUM-4 (raw decode) | `test_raw_register_decode_path_to_blanking` via `from_modbus_registers` (signed 64 + ÷10 108) |
| claude LOW-3 (CLOUD narrowing vs #490) | Documented in DATA_MAPPING + CHANGELOG Unreleased |
| claude LOW-4 (0↔1 flap) | Noted in docstring as accepted cosmetic noise |

Gates: ruff ✓, mypy strict ✓ (46 files), pytest **2835 passed / 3 skipped**.

PR #565, opened 2026-08-12
ci: fix quality-validation compileall target and dead exit-code guards (#549)

### LINKED ISSUE #549: CI: quality-validation.yml compiles non-existent const.py (real package is const/)
Found during a documentation-accuracy audit (verified against the tree, not inferred).

**`.github/workflows/quality-validation.yml`** "Validate Python syntax" step passes `custom_components/eg4_web_monitor/const.py` to `python -m compileall`.

That file **does not exist**. The module is a package: `custom_components/eg4_web_monitor/const/` (containing `brand.py`, `config_keys.py`, `modbus.py`, `sensors/`, ...).

```
$ ls custom_components/eg4_web_monitor/const.py
ABSENT
$ ls -d custom_components/eg4_web_monitor/const
custom_components/eg4_web_monitor/const
```

Two things worth checking while fixing:
1. Point the step at `const/` (or drop the explicit file list and compile the package).
2. The step guards with `if [ $? -eq 0 ]`, which cannot work as intended under the default `bash -e` — a failing `compileall` aborts the step before the check runs. Worth confirming whether this step has ever meaningfully gated.

Related: `CLAUDE.md` also documents `SensorConfig` as living in `const.py`; that doc fix is being handled separately in a documentation-correction PR.

btli:
Fixes #549

## What was broken

**1. The `Validate Python syntax` step listed a file that does not exist — and silently passed anyway.**

The step passed an explicit file list to `python -m compileall`, ending in `.../const.py`. There is no `const.py`: `const/` is a package. Verified locally: `python -m compileall -q <missing file>` prints `Can't list '<path>'` but **exits 0**. So the old step never actually compiled `const.py`, yet passed every run — it under-gated the syntax check, and its file list could silently drift out of date again as modules are added or moved.

Fix: compile the whole tree — `python -m compileall -q custom_components/eg4_web_monitor`. Every target now exists by construction, and new files are picked up automatically instead of requiring the list to be maintained. The compileall is preceded by `test -d custom_components/eg4_web_monitor` because compileall also exits 0 on a nonexistent **directory** argument (same `Can't list`, exit 0) — without the guard, a path rename/typo would silently un-gate the step again, the #549 failure class one level up.

**2. The `if [ $? ... ]` guards were dead code.**

GitHub Actions runs `run:` steps with `bash -e` by default. Under `bash -e`, a command that exits non-zero aborts the script *immediately* — control never reaches the next line, so `if [ $? -eq 0 ]` / `if [ $? -ne 0 ]` after the command can never observe a failure. The guard bodies were unreachable; the step failed (or passed) purely via `bash -e` semantics, and the guards only added noise.

## Each guard fix

- **`Validate Python syntax` (was lines 29–35):** replaced the file list + `if [ $? -eq 0 ]` guard with a single `compileall -q custom_components/eg4_web_monitor` followed by the success echo. On a syntax error, `compileall` exits 1 and `bash -e` fails the step with compileall's own diagnostic; the success echo never runs.
- **`Run ruff linter` (was lines 207–212):** removed `if [ $? -ne 0 ]; ... exit 1; fi`. `ruff check` already prints its findings in GitHub format; under `bash -e` a non-zero exit fails the step directly.
- **`Run ruff formatter check` (same pattern):** same fix — removed the unreachable guard after `ruff format --check .`.
- **`Run mypy type checking` (was lines 831–835):** same fix — removed the unreachable guard after `mypy`. mypy prints each type error itself before exiting non-zero.

No other workflow behavior changed; the success echoes are kept so passing logs look the same.

## Verification

- `python -c "import yaml; yaml.safe_load(open('.github/workflows/quality-validation.yml'))"` → parses clean.
- `python -m compileall -q custom_components/eg4_web_monitor` on the real tree → exit 0.
- `python3 -m compileall -q <missing file>` → prints `Can't list '<path>'`, exit 0 — confirming the old step silently passed while never compiling `const.py`. Same result for a missing **directory**, which is why the step now guards the tree path with `test -d` (verified: `bash -e -c 'test -d <missing dir>; echo should-not-print'` → exit 1, echo suppressed).
- Shell semantics, proven locally with a deliberately broken file (`def broken(:`) in a temp dir:
  - `python3 -m compileall -q <dir>` → exit 1 with a SyntaxError diagnostic.
  - `bash -e -c 'python3 -m compileall -q <dir>; echo OK'` → exit 1, success echo not printed: a broken `.py` fails the step directly.
  - `bash -e -c 'python3 -m compileall -q <dir>; if [ $? -eq 0 ]; then ...; else echo guard-ran; exit 1; fi'` → `guard-ran` **never printed**, demonstrating the old guard body was unreachable (the step's exit 1 came from `bash -e` aborting at `compileall`, not from the guard's `exit 1`).

### DIFF
--- a/.github/workflows/quality-validation.yml
+++ b/.github/workflows/quality-validation.yml
@@ -27,7 +27,12 @@ jobs:
       - name: Validate Python syntax
         run: |
           # Compile the whole tree: compileall exits 0 on a nonexistent file
-          # argument, so an explicit file list can silently skip files (#549).
+          # OR directory argument ("Can't list", success), so an explicit file
+          # list can silently skip files (#549) — and the tree path itself
+          # needs the test -d guard against the same failure class one level
+          # up (rename/typo). Under bash -e both guards fail the step
+          # directly, so no exit-code check is needed.
+          test -d custom_components/eg4_web_monitor
           python -m compileall -q custom_components/eg4_web_monitor
           echo "✅ All Python files have valid syntax"
 
--- a/.github/workflows/quality-validation.yml
+++ b/.github/workflows/quality-validation.yml
@@ -27,11 +27,8 @@ jobs:
       - name: Validate Python syntax
         run: |
           # Compile the whole tree: compileall exits 0 on a nonexistent file
-          # OR directory argument ("Can't list", success), so an explicit file
-          # list can silently skip files (#549) — and the tree path itself
-          # needs the test -d guard against the same failure class one level
-          # up (rename/typo). Under bash -e both guards fail the step
-          # directly, so no exit-code check is needed.
+          # or directory argument ("Can't list"), so an explicit file list can
+          # silently skip files (#549) and the tree path needs the test -d guard.
           test -d custom_components/eg4_web_monitor
           python -m compileall -q custom_components/eg4_web_monitor
           echo "✅ All Python files have valid syntax"
--- a/.github/workflows/quality-validation.yml
+++ b/.github/workflows/quality-validation.yml
@@ -26,13 +26,12 @@ jobs:
 
       - name: Validate Python syntax
         run: |
-          python -m compileall -q custom_components/eg4_web_monitor/__init__.py custom_components/eg4_web_monitor/_config_flow/__init__.py custom_components/eg4_web_monitor/coordinator.py custom_components/eg4_web_monitor/sensor.py custom_components/eg4_web_monitor/number.py custom_components/eg4_web_monitor/switch.py custom_components/eg4_web_monitor/button.py custom_components/eg4_web_monitor/select.py custom_components/eg4_web_monitor/utils.py custom_components/eg4_web_monitor/const.py
-          if [ $? -eq 0 ]; then
-            echo "✅ All Python files have valid syntax"
-          else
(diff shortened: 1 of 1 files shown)


btli:
## Tribunal summary (pollux)

**Verdict: CLEAN** at head `24eb3bf` — ready for human merge.

| Round | Engine | Verdict | Head |
|---|---|---|---|
| 0 | spec-gate (cursor) | SPEC-COMPLIANT | fb59ea0 |
| 1 | codex (blocking) | CLEAN | fb59ea0 |
| 1 | claude_code (blocking) | 1 MEDIUM: compileall exits 0 on missing *directory* too — fixed with `test -d` guard | fb59ea0 |
| 1 | agy (advisory) | CLEAN | fb59ea0 |
| 2 | claude_code | CLEAN | c411b22 |
| 3 | codex (convergence) | CLEAN | c411b22 |
| 3 | agy (advisory) | CLEAN | c411b22 |
| 4 | codex + claude_code (final convergence after dry-pass comment trim) | CLEAN | 24eb3bf |

Dry-pass: 2 runs (comment-only trims). No escalated advisory findings. CI: all checks green.

PR #566, opened 2026-08-12
refactor: remove inert _attr_entity_id dead code (#550)

btli:
## Summary
- Deletes all 17 inert `_attr_entity_id` assignments across `base_entity.py`, `binary_sensor.py`, `button.py`, `select.py`, and `update.py` (option 1 / maintainer-endorsed for #550).
- Removes now-unused `generate_entity_id` imports from those files; leaves `utils.generate_entity_id` in place as a tracked orphan (out of scope).
- Does **not** introduce any `entity_id` / `suggested_object_id` replacement — runtime entity IDs remain HA-derived from slugified names; unique_ids, names, and translation_keys unchanged.
- Updates the three known failing assertion sites to assert real behavior (`unique_id` identity + absence of the dead attribute).

Closes #550.

## Mutation-check
With the production deletion reverted and the updated tests kept, all five modified tests fail on `assert not hasattr(..., "_attr_entity_id")` (attribute present again → `hasattr` is True):

```
FAILED tests/test_operating_state.py::TestOffGridBinarySensor::test_unique_id_and_translation_key
FAILED tests/test_update_entities.py::TestEntityInit::test_no_inert_entity_id_attr_inverter
FAILED tests/test_update_entities.py::TestEntityInit::test_unique_id_stable_across_model_name_noise
FAILED tests/test_update_entities.py::TestEntityInit::test_no_inert_entity_id_attr_gridboss
FAILED tests/test_control_discovery.py::test_parallel_group_unique_ids_include_group_identity
5 failed
```

Assertion lines: `test_operating_state.py:349`, `test_update_entities.py:182/191/200`, `test_control_discovery.py:343`. Unique_id assertions in the parallel-group test would stay green under pure deletion (they assert real identity that never depended on the dead attribute); the `hasattr` guards are what go red.

## Gate evidence
- `grep -rn '_attr_entity_id' custom_components/` → 0 matches
- `uv run ruff check custom_components/ --fix` → All checks passed
- `uv run ruff format custom_components/` → 46 files left unchanged
- `uv run mypy --config-file tests/mypy.ini custom_components/eg4_web_monitor/` → Success: no issues found in 46 source files
- `uv run pytest tests/ -x --tb=short` → **2839 passed**, 3 skipped
- `uv run python tests/validate_silver_tier.py` → 10/10
- `uv run python tests/validate_gold_tier.py` → 6/6
- `uv run python tests/validate_platinum_tier.py` → all platinum checks passed

## Test plan
[...]

Issue #353, opened 2026-07-09
Update to 1E1515 only partial success


eode:
### Integration Version

v3.4.0

### Home Assistant Version

2026.7.1

### Connection Mode

Hybrid (Local + Cloud)

### Device Model(s)

6000XP

### Describe the Bug

When performing an upgrade, instead of the expected 1E1515, the result was 1E1415.  This is also not what the upgrade notifier is expecting, so of course, it stays around.

Also, as a minor issue (probably not related, but perhaps straightforward to fix), the update notification says:
```
Installed version:    ccaa-1E1415
Latest version:    ccaa-1515     # note the missing 1E -- this is confusing, though probably unrelated.
```

Back to the main issue:

The relevant error log is included, but probably will not be useful, as it stems from me attempting to click "update" when the notification returned to its initial state after performing the update.    The error message is, I believe, due to the device being busy with an existing update process.  The error showed up as a popup in the UI.

If, after the device has gone through the *full* update process, I click "update" on the notification again, it attempts to upgrade.  This is reflected by behavior on the device itself, as the device attempts to upgrade.  But, again, this results in 1E1415.

### Steps to Reproduce

1: on the EG4 Web Monitor HomeAssistant update notification, click 'update'.

### Home Assistant Logs

```shell
2026-07-09 15:41:34.184 ERROR (MainThread) [custom_components.eg4_web_monitor.update] Failed to start firmware update for 4413740117: API error (HTTP 200): deviceBusy
2026-07-09 15:41:34.187 ERROR (MainThread) [homeassistant.components.websocket_api.http.connection] [547442894272] Unexpected exception
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/components/websocket_api/commands.py", line 281, in handle_call_service
    response = await hass.services.async_call(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<7 lines>...
    )
    ^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2900, in async_call
    response_data = await coro
                    ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2943, in _execute_service
    return await target(service_call)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 949, in entity_service_call
    response_data = await _async_handle_entity_calls(
[...]

PR #567, opened 2026-08-12
fix: honest indeterminate firmware update progress (#512)

### LINKED ISSUE #512: Firmware update progress jumps 0% → 100% with no intermediate states
Cosmetic follow-up from [#353's beta.4 test report](https://github.com/joyfulhouse/eg4_web_monitor/issues/353#issuecomment): during a real install the update entity shows **Installing (0%)** for the duration and then **Installing (100%)**, with nothing in between, even while the device has visibly reacted (download performed, status light blinking).

The cloud's `remoteUpdate` status endpoint does report intermediate progress for some device/firmware combinations, but on the reporter's 6000XP the observed sequence effectively has two states. Worth investigating whether:
- the endpoint actually emits intermediate percentages for this device and we poll too coarsely / map them away, or
- the device genuinely reports only start/end, in which case an indeterminate progress presentation (no percentage) would be more honest than a frozen 0%.

Low priority — purely cosmetic; the functional multi-step chain behaviour is #353's scope and was fixed in v3.5.1-beta.5.

btli:
## Summary
- Fix #512: while `_install_lock` is held (HA-initiated multi-step firmware chain), `update_percentage` returns `None` so HA shows an indeterminate spinner instead of component-local 0%/100% readings.
- For externally initiated updates (lock not held, row still `in_progress`), map synthetic `0` and stale terminal `100` → `None`; values strictly between 0 and 100 pass through.
- Adds focused property tests around the update-progress region (avoids the `_attr_entity_id` hunks touched by PR #566).

## Mutation check
With the `update_percentage` fix temporarily reverted to the prior pass-through implementation, the three None-mapping tests failed; pass-through / idle cases still passed:

```
FAILED TestProperties::test_update_percentage_none_while_install_lock_held
  assert 100 is None
FAILED TestProperties::test_update_percentage_none_for_external_synthetic_zero
  assert 0 is None
FAILED TestProperties::test_update_percentage_none_for_external_stale_terminal_100
  assert 100 is None
PASSED TestProperties::test_update_percentage_passes_external_intermediate
PASSED TestProperties::test_update_percentage_idle_terminal_100_unchanged
============ 3 failed, 2 passed ============
```

Restored fix → same five tests all green.

## Test plan
- [x] `uv run ruff check custom_components/ --fix && uv run ruff format custom_components/`
- [x] `uv run mypy --config-file tests/mypy.ini custom_components/eg4_web_monitor/`
- [x] `uv run pytest tests/ -x --tb=short` (2844 passed, 3 skipped)
- [x] silver / gold / platinum tier validators
- [ ] Manual: start Install from HA and confirm indeterminate spinner (no Installing 0%/100% flicker across components)
- [ ] Manual: observe an externally started portal update with a mid-range `updateRate` and confirm a real percentage still surfaces

Made with [Cursor](https://cursor.com)


btli:
## Tribunal summary — CLEAN at 66c0470

**Blocking engines:** codex (GPT-5.6-sol) — CLEAN r2; claude_code (claude-fable-5) — CLEAN r1, re-attested r2. Advisory seat unseated (ladder exhausted); round ships on the blocking pair per doctrine.

**Round history:**
- Spec-gate (claude_code): SPEC-COMPLIANT on all 5 contract clauses, pylxpweb citations verified at pin.
- r1: claude_code CLEAN (UX/state-machine lens: HA ≥ 2026.1.0 renders in_progress+None as indeterminate on all supported versions); codex BLOCKING — 2 MEDIUMs.
- Fixes: M1 adjudicated **pre-existing** with git evidence (diff touches only `update_percentage`, zero `in_progress` hunks; wedge timing unchanged, only rendering) → characterization test + limitation documented + #573 filed. M2: central premise split into verified-against-code (pylxpweb ab87902 single-row updateRate parse, no chain aggregation) + asserted-unverified (component-vs-chain portal semantics, #353/#512).
- r2: both engines CLEAN. codex confirmed the M1 disposition sound and #512-scoped; claude_code independently verified ab87902 and the strengthened mutant matrix.

**Follow-ups:** #573 (freshness-bound stale active rows; plus codex r2 LOW — coordinator-succeeds/firmware-poll-fails availability test belongs there).

**Evidence basis:** cloud API + code-path facts; device behavior claims graded (no firmware surface). Mutation checks red-then-green on all new tests.

PR #568, opened 2026-08-12
fix: suppress bogus eps_apparent_power_l1/l2 on hybrid family (#548)

### LINKED ISSUE #548: EG4_HYBRID: eps_apparent_power_l1/l2 are not apparent power (reg 132 is a counter)
## Summary

On **EG4_HYBRID**, input registers 131 and 132 do not carry apparent power, but the integration
exposes them as `eps_apparent_power_l1` / `eps_apparent_power_l2` with unit `VA` and state class
`measurement`.

Register **132 is an incrementing counter** — the same class of defect as #544, where input
register 123 turned out to be an ARM-local 1 Hz counter surfaced as Generator Power.

On **EG4_OFFGRID these same registers are genuine** ([verified here](https://github.com/joyfulhouse/eg4_web_monitor/pull/547)),
so any fix must be family-scoped. This is the mirror image of #544, where the bogus register was
off-grid and genuine on hybrid.

## Evidence

**Firmware** — `FAAB-27xx_20260330_App` (18kPV / FlexBOSS21 ARM comms, sha256 `63d0efba…`, load base
`0x08010000`). FC04 serves input registers from the image at `0x20009DBC` built by `0x080594D8`.

Register 132's source is `0x2000D088 +0/+2`, and its writer increments:

```
0x080154F2  df f8 50 0b   LDR.W R0,[→ 0x2000D088]
            00 88         LDRH  R0,[R0]
            40 1c         ADDS  R0,R0,#1        ; increment
            df f8 48 1b   LDR.W R1,[→ 0x2000D088]
            08 80         STRH  R0,[R1]
```

It advances only after an accumulator at `[+0xA0]` crosses `0x006DDD01` (the `+2` half likewise from
`[+0xA8]` at `0x08015698`). All nine literal-pool instances of `0x2000D088` were audited; every other
write is initialisation, reset, or persistence restore. It is a persistent, threshold-gated event
counter, not an instantaneous measurement.

Register 131 (`0x2000E8C0` / `0x2000E898`) *is* a genuine DSP power quantity, but its writer **splits
it by sign into positive- and negative-direction fields** — negative values are negated into a
separate field and the positive field is zeroed (`0x0804D846`, `0x0804D850`). That is incompatible
with a non-negative apparent power. It is not proven to be apparent power, and not proven to be
anything else.

**Live** — both hybrid inverters on my own system:

```
sensor.18kpv_4512670118_eps_apparent_power_l1         0 VA
sensor.18kpv_4512670118_eps_apparent_power_l2         0 VA
sensor.flexboss21_52842p0581_eps_apparent_power_l1    0 VA
sensor.flexboss21_52842p0581_eps_apparent_power_l2    0 VA
```

They read `0` because EPS output is idle — which is also exactly what a threshold-gated counter reads
before its threshold is crossed. **The live data cannot distinguish the two hypotheses; only the
firmware does.** That is why this has gone unnoticed: it is latent until a hybrid runs sustained EPS
load, after which reg 132 climbs monotonically and never returns to a plausible VA reading, polluting
long-term statistics.

Full trace: `docs/reference/firmware/HYBRID_EPS_REGISTERS.md` (PR #547).

## Proposed fix

Mirror the #544 machinery rather than inventing a new mechanism. Add a hybrid-scoped counterpart to
`OFFGRID_EXCLUDED_SENSORS` in `const/device_types.py`:

```python
# Sensors whose backing registers are NOT measurements on EG4_HYBRID:
# input reg 132 is a threshold-gated counter and reg 131 is a sign-split
# directional power (see docs/reference/firmware/HYBRID_EPS_REGISTERS.md).
# Both ARE genuine on EG4_OFFGRID -- this must stay family-scoped.
HYBRID_EXCLUDED_SENSORS: frozenset[str] = frozenset(
    {
        "eps_apparent_power_l1",
        "eps_apparent_power_l2",
    }
)
```

and a matching branch in `_should_create_sensor` (`sensor.py`) with the **same fail-closed posture** —
an unresolved family (missing features *or* the literal `UNKNOWN` string) creates nothing, since
pylxpweb emits `UNKNOWN` whenever the parameter fetch fails.

Existing entities need purging, as in #544 — the registry cleanup there is the precedent.

### Judgment call for review

Register **132 is certain**; the increment is unambiguous. Register **131 is a weaker case** — it is a
real quantity, just not a non-negative apparent power. Two options:

1. **Exclude both** (proposed). A sign-split directional power published as `VA` is wrong regardless
   of what it is, and reg 131 is the L1 sibling of a sensor we are removing — leaving one half of a
   pair is confusing.
2. **Exclude 132 only.** Narrower blast radius; keeps a value that is at least a real measurement,
   at the cost of shipping something mislabelled.

I lean to (1) but this is worth a second opinion before implementation.

## Out of scope — deliberately

- **Aggregate `eps_apparent_power` (reg 25) stays.** On hybrid it is an ARM-computed `V × I ÷ 100`
  estimate rather than a DSP measurement, but it is a genuine physical estimate. Note its `S ≥ P`
  relation is *firmware-enforced* (clamp at `0x08019ED8`), so that relation cannot be used to
  validate it.
- **`eps_voltage_s` / `eps_voltage_t` (regs 21/22) stay as they are.** They are also unusable —
  live readings of 256.0 V and 4832.0 V on a split-phase system — but they are already suppressed
  everywhere by the `supports_three_phase` gate. Removing them would trade a latent bug for a
  certain regression on three-phase hybrids, which nobody has tested and where those slots may carry
  real voltages. Tracked in the doc, not here. **Do not relax that gate.**

## Validation

- Unit tests for the new gate mirroring the #544 tests: excluded on `EG4_HYBRID`, **created** on
  `EG4_OFFGRID` (the regression risk — these are genuine there), not created on unresolved/`UNKNOWN`.
- Live check on my hybrid plant: the four entities disappear, and off-grid parity is unaffected.
- No pylxpweb change needed — the register definitions are correct as raw registers; this is purely
  about which sensors the integration surfaces per family.

btli:
## Summary

- add a family-scoped `HYBRID_EXCLUDED_SENSORS` contract for `eps_apparent_power_l1` / `eps_apparent_power_l2`
- fail closed while inverter family detection is missing or `UNKNOWN`, while preserving the existing split-phase capability gate and genuine EG4_OFFGRID sensors
- remove already-registered hybrid entities with exact per-serial/device-namespace matching and raise a translated Repairs warning
- keep aggregate `eps_apparent_power`, non-inverter namespaces, and all non-hybrid families unaffected

Fixes #548.

## Validation

- `uv run ruff check custom_components/ --fix && uv run ruff format custom_components/`
- `uv run mypy --config-file tests/mypy.ini custom_components/eg4_web_monitor/`
- `uv run pytest tests/ -x --tb=short`
- `uv run python tests/validate_silver_tier.py`
- `uv run python tests/validate_gold_tier.py`
- `uv run python tests/validate_platinum_tier.py`
- focused #548 / #544 / sensor-entity regression coverage
- translation key and placeholder validation across `strings.json` and every locale file

All gates passed.

## Mutation check

Temporarily reverted both functional behaviors by emptying `HYBRID_EXCLUDED_SENSORS` and disabling the hybrid registry-cleanup invocation, then ran every added/changed test:

```text
uv run pytest tests/test_issue_548_hybrid_eps_apparent_power.py \
  tests/test_sensor_entities.py::TestShouldCreateSensor::test_split_phase_sensor_with_support \
  tests/test_sensor_entities.py::TestShouldCreateSensor::test_default_true_when_feature_flag_missing \
  tests/test_sensor_entities.py::TestCreateInverterSensors::test_no_features_defers_family_sensitive_sensors \
  -q --tb=short
```

Result with the fix reverted: `6 failed` — all three new tests and all three changed legacy tests were RED. After restoring the fix, the focused suites and full test suite passed.


btli:
## Tribunal summary (pollux)

**Verdict: CLEAN** at head `dc607e7`.

| Round | Engine | Verdict |
|---|---|---|
| 0 | spec-gate (cursor) | SPEC-COMPLIANT |
| 1 | codex (blocking) | CLEAN |
| 1 | claude_code (blocking, registry-migration lens) | CLEAN — family fail-closed creation gate, idempotent cleanup, #544 unification behavior-preserving, repairs lifecycle verified |
| 1 | kimi (advisory) | 2 LOW (non-blocking, tracked as follow-ups): restore purge-path docstring rationale in `__init__.py`; ingest new repairs key into llmwiki table |

Dry-pass: 1 substantive (unified #544/#548 suppression + repairs machinery, `dc607e7`) + final no-op. No escalated findings. CI: 34 pass / 1 pending (Silver - Parallel Update Count) — will confirm green before handoff.

btli:
CI note: all check names have successful runs on head `dc607e7`; the one 'pending' shown in the checks rollup is a stale duplicate check-run from a superseded attempt, not a live gate. Ready for merge.

btli:
Hold note: independent byte-level re-verification of the reg 131/132 firmware claims is running before merge (repo rule: re-verify against the primary source; a green check can be wrong). Will post the per-claim verdicts here.

btli:
**Firmware verification complete — hold lifted.**

Independent re-derivation of the FAAB App image (`fw-verify-548-regs`) reproduced the reconstruction byte-identically (351,642 bytes, sha256 `63d0efba...49f4a2`, ARM Thumb, load base 0x08010000) and re-verified every claim this PR rests on:

1. **FC04 table @ 0x20009DBC / builder 0x080594D8 — REPRODUCED** (dispatcher 0x0803F050 → handler 0x0803FCCC; big-endian emission confirmed).
2. **Reg 132 = threshold-gated counter — REPRODUCED** (increment at 0x080154F2 gated on accumulator vs literal 0x006DDD01; +2 mirror at 0x08015698; not a measurement store).
3. **Reg 131 = sign-split — REPRODUCED** (0x0804D838 sign branch negates into a separate negative-direction field and zeros the positive field; incompatible with non-negative apparent power).
4. **No live-measurement writer to 0x2000D088 — REPRODUCED** (all 9 literal-pool instances + 80 PC-relative refs audited; every store is counter/init/persistence/reset; movw/movt sweep clean). Stated frame caveat: an arithmetically-computed address would evade both audits.

The exclusion of hybrid `eps_apparent_power_l1/l2` is justified by the bytes, not just the cited document. This PR is ready to merge.

Issue #310, opened 2026-07-03
Switches don't seed the parameter cache after cloud-fallback writes (link-down stale-revert, #301 class)


btli:
Found during the PR #306 review (Codex adversarial pass). Pre-existing, cross-cutting — affects the whole working-mode switch family (Charge Last, Green/Off-Grid Mode, EPS, Quick Charge, Fast Zero Export, Share Battery), not any single PR.

PR #301 gave number/select/time entities `async_write_with_cloud_fallback` + `note_parameters_written()`: after a cloud-fallback write under HYBRID link-down, the acknowledged values are seeded into the parameter cache so the entity converges on what was written. Switches never migrated: `_execute_local_with_fallback` → `_execute_cloud_function_action` (base_entity.py:1209-1293) sleeps, calls `async_refresh_device_parameters`, and clears optimistic state — but under link-down `_refresh_device_parameters` (coordinator_mixins.py:2399) skips the read, and its docstring's promise ("entities converge on written values via note_parameters_written()") is one switch.py doesn't keep. A switch flipped while the local link is down can revert to the stale pre-write value once optimistic state clears, until link recovery.

The gap is untested for the whole switch family: `test_hybrid_local_fail_falls_back_to_cloud` covers local-raises→cloud-succeeds, not cloud-succeeds→refresh-skipped→stale-revert.

Fix shape: migrate `_execute_cloud_function_action` to seed via `note_parameters_written({param_key: bool})` when the write landed via cloud with a transport attached (mirror utils.async_write_with_cloud_fallback), + one link-down regression test per pattern. Also worth folding in: the `enabled_default is False` identity check in switch.py should be truthiness-based (`not mode_config.get("enabled_default", True)`) so a future non-bool value can't silently ship an entity enabled.

btli:
[...]

PR #586, opened 2026-08-15
feat: add immutable raw snapshot store

### LINKED ISSUE #583: Phase A3: immutable raw snapshot store and freshness policy
## Goal

Implement planning item 3 from `llmwiki/40-hardware/dongle-emulation.md`: a bounded immutable raw-register snapshot store fed exclusively by the released `pylxpweb 0.10.0b3` observer seam.

## Binding architecture

- Pin the public PyPI prerelease `pylxpweb 0.10.0b3`; no Git/private pin. The b3 bump is required by the reviewed public observer-detach/control seam used for coverage loss and terminal teardown; b2's construction-only observer injection cannot satisfy those lifecycle requirements.
- Create the store only for A2 `ELIGIBLE` LOCAL/HYBRID owners with complete direct `modbus_tcp`/`modbus_serial` coverage; all existing fail-closed eligibility results remain blocking.
- The construction-time observer stages exact terminal-winning raw segments inside one owner refresh. Coordinator/owner lifecycle brackets the cycle.
- Publish atomically only if the entire refresh succeeds, every invoked read contributed, transport/coverage remained valid, and the public monotonic observer-error counter did not increase. Otherwise discard the builder, increment a redacted suppressed-incomplete count, and retain the prior complete frame unchanged.
- Do not invent or require an upstream incomplete-range event; `0.10.0b3` has no such API.
- Exactly one externally readable latest-complete frame per `(endpoint, unit)`; mutable staging is private and never readable. O(1) storage; no ring buffer or latest-attempt raw store.
- Frame identity is `(owner_epoch, generation)`; generation/poll-cycle increase within one owner epoch. Reload creates a distinct epoch and may restart generation. Old stores are inaccessible after unload.
- Immutable block/frame fields: endpoint alias/key, unit, family/firmware scope, register space/start/count/exact word tuple, epoch, generation, poll-cycle, monotonic start/end, and honest validation/CRC state (`not_applicable` for TCP, `valid` for RTU). No parsed reconstruction.
- One named freshness policy: `min(60s, max(15s, 3 * configured local poll interval))`, using monotonic acquisition end only and a `None` never-ran sentinel.
- Store-on versus store-off produces identical Modbus request count and order; zero additional periodic or reactive reads.
- No emitter, cloud parser/connection, listener/socket/port 8000, writes, identity credentials, HA entity-state ingestion, portal response ingestion, or parsed/scaled model ingestion.
- Unload/cancellation/coverage loss publishes nothing partial, removes eligible lookup, releases retained raw data, and leaves no task/socket/callback.
- Redacted bounded metrics only: freshness/age, suppressed incomplete, observer failures/health; no serials, raw words, endpoint identity labels, or unbounded error text.

## Acceptance

- Dependency/lock/manifest consistency for `0.10.0b3`.
- Frozen/slotted immutable models and atomic single-writer publication.
- Mutation-proven tests for complete publication, failed/incomplete suppression, observer-error demotion, stale boundary, wall-clock immunity, epoch/reload behavior, LOCAL/HYBRID eligibility, zero extra reads/order, no cloud/entity/parsed import path, cancellation/unload, O(1) retention and redacted representations/metrics.
- Every new/materially changed test demonstrates RED with its invariant removed and GREEN restored.
- Full repository tests, Ruff, format, configured strict mypy, translations/architecture gates, current-head CI.
- Public-repo draft PR, DCO/co-author trailers, no merge until DRY and dual blocking tribunal CLEAN.

## Excluded

Cloud framing/emission, reactive reads, listeners, vendor identity/session handling, controls, OTA, HIL, and package publication.

btli:
Closes #583

## Summary

- pin the public PyPI `pylxpweb==0.10.0b3` release on the repository canonical dependency surfaces
- add the A2-owner-scoped frozen latest-complete raw snapshot model/store with epoch/generation identity and monotonic freshness
- stage exact public register-observer segments privately and publish only after an entirely successful eligible LOCAL/HYBRID owner refresh
- suppress failed, cancelled, incomplete, observer-error, overflow, and coverage-loss attempts while retaining the prior complete generation
- clear callback lookup and retained raw state on coverage loss/unload without adding reads, tasks, sockets, listeners, or alternate ingestion paths

## r3 remediation (60fd489)

- raw observer capability is now represented independently of initial eligibility, so a capability created while ineligible can be enabled when coverage later becomes eligible (production-path transition covered by `test_initially_ineligible_capability_enables_on_later_eligibility`)
- observer attach is transactional: `add()` and coverage restore commit no record, state, or enabled flag until attach succeeds, and failed coverage remains retryable/disabled
- coverage loss quarantines local snapshot state before the best-effort observer detach, so stale frames can never remain visible after a failed detach; per-capability detach failures are aggregated after every capability is processed
- the per-historical-unit identity counter dict is replaced by one bounded owner high-water mark; `(owner_epoch, generation)` still never repeats across coverage toggles
- `RawRegisterBlock.words` / `RawSnapshotFrame.blocks` now require exact tuples with exact element types at runtime
- `complete_snapshot_refresh` is annotated `contextlib.AbstractAsyncContextManager[None]` (typing only, no runtime change)

## Dependency consistency evidence (review finding 7)

Dependency surfaces were **not** modified by the r3 remediation. Git-object proof at head `60fd489416b055bc8b992be0fe823dbc97464b86`:

- `git rev-parse HEAD:custom_components/eg4_web_monitor/manifest.json` → `9181d28e3e94db6eecacad67607151e36fb51fe0` (contains `pylxpweb==0.10.0b3`)
- `git rev-parse HEAD:tests/requirements-test.txt` → `307bbc748cc5406b3902cb662891c76a773a63a9` (contains `pylxpweb==0.10.0b3  # keep in sync with manifest.json`)
- `git diff 05612f106b63e44dc42373c586b0cc1d8276754f..HEAD -- manifest.json tests/requirements-test.txt` → empty (blob SHAs identical to the r3 review base)

## Verification

- `python -m pytest tests/ --ignore=tests/test_plant_api.py` — 3156 passed, 9 skipped
- focused snapshot/owner/coordinator set (`test_endpoint_raw_snapshot.py`, `test_raw_snapshot.py`, `test_endpoint_bus_owner.py`, `test_coordinator_local.py`) — 240 passed
- `python -m ruff check .` — passed
- `python -m ruff format --check .` — passed (143 files)
- strict configured mypy (`mypy --config-file tests/mypy.ini custom_components/eg4_web_monitor/`) — passed (48 source files)
- translations and Silver/Gold/Platinum validators — passed

## TDD / mutation proof

Each added/materially changed invariant was first observed RED, then GREEN. For the r3 round: all 8 new tests fail with `custom_components/eg4_web_monitor/endpoint_bus.py` + `raw_snapshot.py` stashed (`8 failed`) and pass with the fix restored (`8 passed`). Mutations covered initial-eligibility transition, attach transactionality on construction and restore, detach ordering and best-effort multi-capability release, bounded identity retention, and exact tuple/element container validation. Earlier rounds covered missing model/module, absent construction observer/lifecycle, publication guards, cancellation after a completed segment, stale-boundary comparison, wall-clock clock source, fail-closed eligibility, extra reads/order, epoch/store/unit/CRC identity, generation monotonicity, forbidden import routes, overflow/resource bounds, and bounded health counters.


btli:
## Base sync: main @ v3.5.1-beta.12 (pylxpweb 0.10.0b4)

**Head:** `412e9263` → `fdcb58ed` (merge of `origin/main` @ `e9853eb`, conflict-free — only `CHANGELOG.md`, `manifest.json`, `tests/requirements-test.txt` came in from main).

**Pin ride-along:** the branch now carries main's `pylxpweb==0.10.0b4` pin (manifest + requirements-test). Issue #583 named 0.10.0b3, but the binding requirement is the public observer seam introduced *in* b3; b4 only adds the dead-link probe fix (pylxpweb #310). Verified against the installed PyPI `0.10.0b4` wheel that every seam symbol this PR consumes exists with the expected shape:
- `RegisterObserverControl.set_register_observer(observer: RegisterObserver | None) -> None`
- `RegisterObservation(register_space, segments)`, `RegisterSegment(start_address, words)`, `RegisterSpace.{INPUT,HOLDING}`, `create_transport_from_config`

No hardcoded `0.10.0b3` references remain in the six Phase A3 deliverable files (the two repo-wide mentions left are historical: the beta.12 CHANGELOG narrative and an accurate `pylxpweb>=0.10.0b3` behavior comment in `tests/test_coordinator.py`).

**Gates at `fdcb58ed`** (ruff 0.15.5, mypy 2.3.0, Python 3.13, deps from `tests/requirements-test.txt`):
- ruff check + format: clean, no diffs
- mypy `--config-file tests/mypy.ini`: Success, 48 source files
- pytest: **3159 passed, 9 skipped**
- Silver / Gold / Platinum tier validators: all pass

Staying DRAFT pending the final adversarial review round.

btli:
## Adversarial round 1 adjudication (Codex gpt-5.6-sol + Kimi K3) — head `fdcb58ed` → `a5d97ffd`

All findings adjudicated against the code and the installed PyPI `pylxpweb 0.10.0b4` wheel. **All confirmed; none refuted.** Every fix carries a regression test demonstrated RED with the fix reverted (`git stash` of the production files) and GREEN restored.

| # | Reviewer / Sev | Finding | Verdict | Fix | Regression test |
|---|---|---|---|---|---|
| 1 | Codex HIGH | HYBRID polls via `_refresh_station_devices()` without a snapshot-refresh context → observer callbacks discarded, `latest_complete_snapshot` permanently `None` | **Confirmed.** Spec #583 explicitly scopes the store to "A2 ELIGIBLE LOCAL/**HYBRID** owners" and `_create_bus_capability` enables snapshots for HYBRID; `observe()` provably drops everything outside an active attempt. | `coordinator_http.py`: the sequential endpoint-group loop now brackets each owner-issued capability refresh — publish on success; abort on failure, cloud-fallback cycle, or known coverage loss (mirrors the LOCAL path) | `test_hybrid_refresh_brackets_snapshot_and_publishes`, `test_hybrid_failed_refresh_aborts_and_retains_prior_complete` |
| 2 | Kimi HIGH | `read_battery` → `None` still emits observations in 0.10.0b4, so the set-equality completeness check suppresses every poll of a battery-less LOCAL inverter | **Confirmed against the wheel:** `_register_data.py` `read_battery` notifies observed segments *before* `return result` on the battery-less SUCCESS path (only the swallowed-failure/BMS-short path returns early without emitting). The covering test's fake was emission-unfaithful, which is why it stayed green. | `endpoint_bus.py`: completeness is now a superset check — every REQUIRED invocation observed; a not-required-yet-observed battery invocation counts as contributed and its raw blocks publish. Test fake's `None` path made emission-faithful (both wheel paths modeled via `battery_none_emits`) | `test_zero_observation_battery_none_does_not_hide_complete_reads` (rewritten, covers both wheel None-paths + required-read suppression) |
| 3 | Codex MED | Per-device publish before entry-wide coverage recompute — a frame can publish after a sibling endpoint's link-down is already declared, readable until end-of-cycle `_sync_transport_link_state` clears stores | **Confirmed** (concurrent endpoint groups + sequential detection make the window seconds long). | Publish-time coverage gate `_snapshot_coverage_known_lost()` at snapshot-context exit in BOTH the LOCAL and HYBRID brackets: positively known degradation (tracked device with no transport or declared link-down) aborts the publish. Deliberately weaker than full eligibility so first-cycle boot ordering can't false-abort; the end-of-cycle recompute remains the authoritative fail-closed check. Residual: frames from *earlier* cycles stay readable until that recompute — unavoidable (detection-bounded) and freshness-policy-bounded. | `test_local_publish_gated_when_sibling_endpoint_known_link_down` |
| 4 | Codex MED / Kimi 2 | `begin_shutdown_capabilities()` raise aborts unload before terminal disconnect + cleanup | **Confirmed** — the real caller chain (`async_unload_entry` → `coordinator.async_shutdown` → `_disconnect_all_transports`) runs bare; a raise skipped terminal disconnect, listener removal, background-task cancel, the base-class shutdown AND the HTTP-client close. A pre-existing test (`test_ha_registry_retries_failed_unload_after_coordinator_disposal`) had pinned the defective propagation and was updated to the containment contract. | `coordinator_mixins.py`: both hops contained (admission-close raise → terminal shutdown retries the detach; terminal-failure group → capabilities retained for `async_retry_failed_shutdowns`); caller cancellation is NOT swallowed (drain re-raises bare `CancelledError`). Kimi's extra site: same containment around `set_snapshot_coverage` in `_sync_transport_link_state` (unreachable with the pinned wheel; consistency hardening). | `test_unload_survives_rejected_observer_detach_and_stays_retryable`, `test_link_state_sync_contains_rejected_coverage_detach`, updated `test_ha_registry_retries_failed_unload_after_coordinator_disposal` |
| 5 | Codex LOW / Kimi 3 | `owner.add()` failure strands a newly created empty owner in `registry._owners` | **Confirmed** — worse than a leak: the stale key makes `create_discovery_capability` raise `EndpointOwnerInUseError`, blocking config-flow discovery for that endpoint until HA restart. | `endpoint_bus.py` `create_capability`: factory AND add failures now pop a just-created record-less owner (pre-existing owners with live records untouched). | `test_failed_observer_attach_during_add_retains_no_record_or_state` (updated: owner removed + discovery unblocked) |

**Gates at `a5d97ffd`** (ruff 0.15.5 isolated-config, mypy 2.3.0 strict, Python 3.13, PyPI wheel deps):
- ruff check + format: clean
- mypy `tests/mypy.ini`: Success, 48 files
- pytest: **3164 passed, 9 skipped** (was 3159 — 5 new regression tests, 1 rewritten in place)
- Silver / Gold / Platinum validators: all pass

Grok's round-1 output was garbled/incomplete — treated as no verdict; a fresh Grok reviews this head in round 2. Staying DRAFT.

btli:
## Adversarial round 2 adjudication (Codex gpt-5.6-sol; Grok CLEAN) — head `a5d97ffd` → `403b1b33`

All three findings confirmed against the code and the installed PyPI `pylxpweb 0.10.0b4` wheel; none refuted. Kimi round 2 pending.

| # | Sev | Finding | Verdict | Fix | Regression test |
|---|---|---|---|---|---|
| 1 | HIGH | Round-1 completeness fix overcorrected: a `read_battery()` → `None` was marked non-required even when it emitted **nothing**, so the wheel's swallowed-failure (BMS timeout/short) None path published the other observations as a fresh "complete" frame despite one invoked read contributing nothing | **Confirmed.** The reviewer's proposed discriminator holds exactly in the wheel: `_register_data.read_battery`'s battery-less SUCCESS path notifies its observed segments *before* `return result`; the failure path (`bms_ok` False) returns *before* the notify — so emission, not the result value, distinguishes the two None paths. (This half of the defect predates round 1: the original set-equality check also published the zero-emission None path — the pre-existing covering test pinned it with an emission-unfaithful fake.) | `endpoint_bus.py`: completeness now requires **every completed invocation to be observed** (`set(observed) == set(invocations)`); the result-based battery special-case marking is deleted. Safety of the stricter rule verified per method against the wheel: all 12 capability-exposed read methods notify on success, so no other read can complete unobserved | `test_battery_none_paths_split_by_emission_publish_vs_suppress` — three blocks: emission-None publishes both raw blocks (RED under pre-round-1 code), zero-emission-None suppresses + retains prior (RED under round-1 code — demonstrated: round-1 published a fresh generation-2 frame), required-read-no-emission suppresses |
| 2 | MED | Coverage gate ignored configured-but-uncreated devices, so a healthy endpoint could publish during the FIRST LOCAL poll while a sibling's connection ultimately failed | **Confirmed.** | `coordinator_local.py`: gate renamed `_snapshot_coverage_unresolved` — a frame publishes only while **every configured direct serial is tracked, transport-attached, and link-up**; a configured-but-untracked sibling now blocks. First-boot cost bounded: single-endpoint entries unaffected (own device is cached before its context exits); multi-endpoint entries publish from the first cycle where all devices exist (at most the boot cycle's frames lost). End-of-cycle eligibility recompute stays authoritative | `test_publish_blocked_until_all_configured_endpoints_resolve` — RED demonstrated by reverting only the gate hunk (published while sibling unresolved), GREEN restored; also asserts the entry publishes once the sibling resolves |
| 3 | MED | Round-1 containment swallowed shutdown failures while the capability stayed retained: a persistent detach rejection let unload return with owner/raw (callback+socket) alive indefinitely unless the endpoint was later re-attached | **Confirmed** (`async_retry_failed_shutdowns` only runs on re-attach to the same endpoint — after unload that may never happen). | `endpoint_bus.py` `owner.shutdown`: the detach rejection is captured, the **terminal close still runs and releases everything** (record dropped, owner closes, raw disconnected+dropped), and the failure is swallowed-with-warning only *after* release — the stale callback on the dropped raw is inert (`observe()` returns on a missing record). `capability.async_shutdown`'s sync `begin_shutdown` pre-call gets the same containment. A **failed terminal close** (socket genuinely still open) still propagates and stays retained for the registry retry — swallow the exception, not the cleanup | `test_failed_detach_during_shutdown_still_releases_every_capability` + updated `test_unload_survives_rejected_observer_detach_and_releases_resources` (both RED under round-1 code: raise / retained-with-socket); `test_terminal_close_failure_stays_retained_and_detach_retries` pins the terminal-failure retained contract that now exclusively owns retry semantics (replaces `test_terminal_completion_waits_for_successful_observer_detach`, which pinned the defective wait-for-detach behavior) |

**Gates at `403b1b33`** (ruff 0.15.5 isolated-config, mypy 2.3.0 strict, Python 3.13, PyPI wheel deps):
- ruff check + format: clean
- mypy `tests/mypy.ini`: Success, 48 files
- pytest: **3165 passed, 9 skipped**
- Silver / Gold / Platinum validators: all pass

Staying DRAFT pending Kimi round 2 on this head.

btli:
## Adversarial round 3 adjudication (Codex gpt-5.6-sol; Grok CLEAN) — head `403b1b33` → `d762a3b7`

One MED finding; **confirmed**, none refuted.

| # | Sev | Finding | Verdict | Fix | Regression test |
|---|---|---|---|---|---|
| 1 | MED | `_disconnect_all_transports()` swallows terminal-close failures while the capability stays retained — post-unload there is no retry path (`async_retry_failed_shutdowns` runs only when the same endpoint is set up again), so a failed Modbus/serial close leaves the owner tombstoned in the **HA-scoped** registry with a possibly open socket/callback indefinitely | **Confirmed** — the coherent completion of the round-1/round-2 contract: retention-for-retry only makes sense while the entry stays loaded. | New `EndpointBusRegistry.force_release_capabilities()`: at unload, after the best-effort close fails, every registry record is unconditionally dropped — record + owner released, failed-set entry discarded, no owner lookup outlives the entry, a later setup of the same endpoint is no longer blocked by the tombstone. The stale observer callback on the dropped raw is inert (`observe()` returns on a missing record) and the abandoned transport's socket closes with GC. **Registry-level retained/retry semantics are unchanged** for loaded-entry paths — `test_terminal_close_failure_stays_retained_and_detach_retries` still pins them. Cancellation is still never swallowed. | `test_unload_force_releases_capability_after_failed_terminal_close` (asserts full release AND that a fresh setup on the endpoint succeeds) + updated `test_ha_registry_retries_failed_unload_after_coordinator_disposal`; both demonstrated RED under the round-2 code (owner tombstoned, capability retained, `EndpointOwnerClosingError` on re-setup) and GREEN with the fix |

Also removed a stray inert `# type: ignore[method-assign]` left in a round-2 test (tests are not mypy-scanned; repo policy forbids suppression comments).

**Gates at `d762a3b7`** (ruff 0.15.5 isolated-config, mypy 2.3.0 strict, Python 3.13, PyPI wheel deps):
- ruff check + format: clean
- mypy `tests/mypy.ini`: Success, 48 files
- pytest: **3166 passed, 9 skipped**
- Silver / Gold / Platinum validators: all pass

Staying DRAFT pending Kimi rounds 2/3 verdicts.

btli:
## Adversarial round 4 adjudication (Codex gpt-5.6-sol; Grok CLEAN) — head `d762a3b7` → `e44c8362`

Finding 1 **confirmed and fixed**; finding 2 **adjudicated: publish-time validity is the spec's contract — the post-publish window is documented and pinned, not patched with a staged commit** (evidence below).

| # | Sev | Finding | Verdict | Resolution | Test |
|---|---|---|---|---|---|
| 1 | MED | `force_release_capabilities()` never restores a nonempty shared owner: A's failed terminal close flips the owner to CLOSING (`owner.shutdown`'s failure branch), force-release pops A's record, and a sharer B's live record is stranded behind `EndpointOwnerClosingError` permanently | **Confirmed** by state-machine inspection — the CLOSING flip was previously temporary-until-retry; round-3's record drop made it permanent for survivors. | When records remain after the drop, owner state is recomputed exactly like a successful per-record shutdown (`CLOSING` only while a survivor is closing, else `OPEN`). Survivors' raw transports are never touched (force-release only drops the released record's raw); the owner still fully closes + deregisters when the released record was the last | `test_force_release_restores_shared_owner_for_surviving_records` — two capabilities sharing an endpoint; A unloads with failing terminal close; B keeps reading AND publishing, and the endpoint accepts a new capability. RED under round-3 code (survivor stranded), GREEN with fix |
| 2 | MED | Post-publish sibling link-down: A's validly published frame stays readable until end-of-cycle coverage sync — the round-1/2 residual window called out as still-open | **Adjudicated, option (b).** Spec evidence: issue #583 line 9 scopes the publication predicate to "**inside one owner refresh**. Coordinator/owner lifecycle brackets the cycle", so line 10's "transport/coverage remained valid" is a *through-publish-time* requirement — enforced by the round-2/round-4 gate. Line 18's coverage-loss requirement ("removes eligible lookup, releases retained raw data") is discharged at the end-of-cycle sync. Decisive against option (a): **a staged end-of-cycle commit cannot remove the window, only move it** — a link that physically dies right after the staged commit leaves the committed frame readable for one full detection latency regardless; detection-bounded exposure is irreducible in any polling system. Staging would also retain a second frame per unit transiently, in tension with the spec's "exactly one externally readable latest-complete frame per (endpoint, unit); O(1) storage". The exposure is bounded by one update cycle (sync) and by the freshness policy for `latest_fresh_snapshot` consumers | `test_coverage_loss_after_publish_retro_invalidates_at_cycle_sync` pins **both halves of the contract explicitly**: (1) the frame remains readable in the post-publish window (asserting the documented behavior so any accidental "fix" trips it), (2) `_sync_transport_link_state` retro-invalidates — lookup `None`, store count 0. Contract also documented in the `_snapshot_coverage_unresolved` docstring |

**Gates at `e44c8362`** (ruff 0.15.5 isolated-config, mypy 2.3.0 strict, Python 3.13, PyPI wheel deps):
- ruff check + format: clean
- mypy `tests/mypy.ini`: Success, 48 files
- pytest: **3168 passed, 9 skipped**
- Silver / Gold / Platinum validators: all pass

Staying DRAFT pending outstanding Kimi verdicts.

PR #579, opened 2026-08-14
feat: add safe dongle capture sanitizer and streaming decoder

btli:
Implements Phase A1 from `llmwiki/40-hardware/dongle-emulation.md` (Phase A, slice 1).

Closes #576.
Bead: `eg4-11ac`.

## What changed

- reassembles bounded TCP streams before frame parsing or redaction
- handles fragmentation, coalescing, duplicate/overlapping segments, sequence wrap, reconnect epochs, truncation, oversize frames, deadlines, and resource ceilings
- fails closed on malformed input, CRC, identity, function, range, EOF, timeout, bounded input/output failures, and unknown schema behavior
- emits only a minimal synthetic JSON fixture from explicitly authorized offline input
- removes raw identities, endpoints, payloads, register values, timestamps, and paths from output and errors
- adds no live network behavior, cloud semantics, listener, bus owner, controls, or runtime integration

## Declared write scope

- `scripts/decode_cloud_frames.py`
- `tests/fixtures/dongle_emulation/minimal_sanitized.json`
- `tests/requirements-test.txt`
- `tests/test_decode_cloud_frames.py`

## Security review

- no capture or binary artifact is tracked
- full ten-commit Gitleaks scan: zero findings
- diagnostics-key scan and quoted ten-character literal scan completed over the full diff
- every matched identity is explicitly synthetic/canary data; addresses are documentation/test-only; the test MAC is locally administered
- no production identity, credential, endpoint, address, frame, or register value is present
- no live environment or vendor cloud was accessed

## Mutation RED/GREEN evidence

All mutations were temporary and restored from `HEAD` before publication.

1. Central bounded-policy hunk replaced with unconditional rejection, then:

   `uv run pytest tests/test_decode_cloud_frames.py -q --tb=no`

   RED: 80 failed, 5 passed. The failing node IDs covered every parser-policy, stream decoder, TCP reassembly, sanitizer, PCAP adapter, redaction/schema, identity/function/range, EOF/timeout, capacity, reconnect, and normal output-publication case. The only passing cases were the three parameterized `test_public_crc_seam_normalizes_bytes_like_inputs` cases, `test_crc_seam_loads_without_pymodbus`, and `test_output_failures_remove_temporary_and_published_files`.

2. CRC implementation hunk replaced with a zero result, then:

[...]

Issue #576, opened 2026-08-14
Phase A1: safe dongle capture sanitizer and streaming decoder


btli:
Implements slice 1 of `llmwiki/40-hardware/dongle-emulation.md`'s Phase A plan.

Scope:
- Upgrade the offline decoder to perform bounded TCP stream reassembly before frame parsing/redaction.
- Produce only synthetic, minimal golden fixtures from authorized offline inputs.
- Add fragmentation, coalescing, duplicate-segment, truncation, oversize, CRC, identity/function/range, EOF, timeout, and redaction tests where supported by captured evidence.
- Preserve unknown behavior as fail-closed; do not invent live cloud semantics.

Security:
- No production serial, PIN/check code, address, MAC, token, capture, frame, or register value in commits, fixtures, issue/PR text, logs, or diagnostics.
- Do not access the live environment or vendor cloud for this slice.

Acceptance:
- Sanitizer performs TCP stream reassembly before redaction.
- Generated fixtures contain only synthetic identities/addresses and minimum assertion data.
- Structural secret scan and literal-free mutation red/green evidence pass.
- Relevant tests, Ruff, strict mypy, full CI, DRY pass, and tribunal are green.
- Draft PR references this issue and the canonical spec.

btli:
Phase A1 draft PR #577 is green at head `157a2ed1` (focused tests, Ruff, configured strict mypy, and 35/35 CI), but it is **not mergeable**. Tribunal round 3 produced blocking-engine delivery failures: the large artifact was stripped on initial delivery, and both permitted retries failed before either engine could inspect code. The three-round safety valve is now active. Human authorization is required before an exceptional transport-only re-review using chunked plain-text delivery; no merge or Phase A2 work will proceed as if Phase A1 were CLEAN.

btli:
Round 5 on head `24269355`: Codex blocking review is CLEAN; Claude Fable exhausted usage credits before review, so the required second blocking attestation is unavailable. Local focused tests/lint/type gates are green. PR #577 remains draft and not mergeable; no model substitution was made.

PR #598, opened 2026-08-29
fix: fail-closed protocol-version/heartbeat strictness in dongle sanitizer

### LINKED ISSUE #588: Optional: fail-closed protocol-version/heartbeat strictness in dongle sanitizer
Follow-up from the #579 vs #577 sanitizer comparison (closed #579 as superseded by #577+#585).

Main's `_sanitize_frame` (scripts/decode_cloud_frames.py ~L846-858) accepts any protocol version / address byte and any non-empty heartbeat payload, masking to `SYNTHETIC_STATUS`. Output stays fully synthetic (no leak), but #576's scope ("preserve unknown behavior as fail-closed") arguably wants stricter rejection.

Proposed (~5 lines, from #579's implementation): reject frames whose protocol version != 0x0001 or address byte != 1 with a new `PROTOCOL` FailureReason, and require heartbeat payload to be exactly one byte. Non-security hardening; low priority. Every other property in #579 is already equal-or-better on main.

btli:
Fixes #588

## What changed

`_sanitize_frame` in `scripts/decode_cloud_frames.py` accepted any protocol version / address byte and any non-empty heartbeat payload, masking unknown variants into known-looking `SYNTHETIC_STATUS` records. Per #576's fail-closed scope, unknown shapes are now rejected instead, using the implementation from closed PR #579:

- Frames whose protocol version is not `0x0001` or whose address byte is not `1` are rejected with a new `PROTOCOL` `FailureReason`.
- The heartbeat (`0xC1`) payload must be exactly one status byte; anything else is rejected (`PROTOCOL` for longer payloads, `MALFORMED` for the zero-byte case, which the minimum-frame-length check catches first).

Output remains fully synthetic in all cases; this only narrows what is accepted.

## Tests

New in `tests/test_decode_cloud_frames.py`:

- `test_sanitizer_accepts_baseline_protocol_frame` — accepted baseline frame still passes
- `test_sanitizer_rejects_unsupported_header_and_heartbeat_shapes` (parametrized) — wrong protocol version, wrong address byte, and two-byte heartbeat payload all rejected with `PROTOCOL`
- `test_sanitizer_rejects_empty_heartbeat_payload` — zero-byte heartbeat payload rejected (`MALFORMED`)

The two canary redaction tests (`test_sanitizer_emits_only_synthetic_minimal_data`, `test_offline_pcap_adapter_and_cli_never_report_source_metadata`) previously carried the canary blob through the now-strict heartbeat payload; they now carry it through data-read-response register words via a `_canary_register_words()` helper, preserving full canary redaction coverage on the accepted path.

## Gates

- Ruff 0.15.5 (CI pin) check + format: pass
- mypy (`tests/mypy.ini`) on `custom_components/eg4_web_monitor/`: no issues in 47 source files
- Full pytest: 3,112 passed, 9 skipped
- Silver / Gold / Platinum validators: pass

### DIFF
--- a/scripts/decode_cloud_frames.py
+++ b/scripts/decode_cloud_frames.py
@@ -171,6 +171,7 @@ class FailureReason(StrEnum):
     OUTPUT_EXISTS = "output_exists"
     OVERSIZE = "oversize"
     PREFIX = "prefix"
+    PROTOCOL = "protocol"
     RANGE = "range"
     SCHEMA = "schema"
     TIMEOUT = "timeout"
@@ -845,6 +846,11 @@ def _sanitize_frame(
 ) -> dict[str, Any]:
     if len(frame) < 19 or 6 + int.from_bytes(frame[4:6], "little") != len(frame):
         raise CaptureError(FailureReason.MALFORMED)
+    # Phase A1 capture evidence contains only protocol version 0x0001 and
+    # address 1. Unknown variants are rejected rather than redacted into a
+    # known-looking record.
+    if frame[2:4] != b"\x00\x01" or frame[6] != 1:
+        raise CaptureError(FailureReason.PROTOCOL)
     session.outer_identity = _bind_identity(session.outer_identity, frame[8:18])
     base: dict[str, Any] = {
         "direction": direction,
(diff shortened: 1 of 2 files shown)


btli:
Adversarial review round 1 finding (protocol-version byte order) adjudicated against the primary source and **confirmed**: `pylxpweb` `transports/dongle.py` defines `PROTOCOL_VERSION = 1` and packs it with `struct.pack("<H", ...)`, so legitimate frames carry `A1 1A 01 00 ...` — the check merged from #579 (`b"\x00\x01"`) and the `_cloud_frame` test helper both pinned the inverted encoding and would have rejected every real version-1 frame.

Fixed in 75eb641:
- sanitizer accepts `frame[2:4] == b"\x01\x00"` (address-byte offset/value were already correct);
- `_cloud_frame` emits the byte-faithful wire encoding, so baseline-accept, golden-fixture, and pcap tests all pin it;
- reject cases now use `02 00` (version 2) and `00 01` (the inverted encoding), both `PROTOCOL`;
- two decoder-level tests hardcoding `00 01` updated for byte-faithfulness.

Mutation check: re-inverting the comparison fails 25 tests. Full gates re-run: ruff 0.15.5 check+format clean, mypy clean (47 files), pytest 3,113 passed / 9 skipped, Silver/Gold/Platinum validators pass.

PR #569, opened 2026-08-12
fix: route unverified off-grid writes cloud-only; stop doomed pure-LOCAL H233 quick-charge write (#558)

btli:
Fixes #558 (option 2 — the #471/#472 pattern for unpinned/unverified writes). Incorporates spec-gate round (regs 158/159) and tribunal round 1 (fail-closed polarity, reg 66).

## Task A — protected registers route CLOUD-ONLY unless the family is positively resolved non-off-grid

**Protected registers**: AC-charge SOC window **160/161**, AC-charge voltage window **158/159**, and AC charge power **66** (tribunal round 1). Local writes to all of them are hardware-UNVERIFIED on the off-grid family:

- **160/161**: all #331 write evidence is the cloud holdParam path.
- **158/159**: only write evidence is a cloud-path delta-test — llmwiki grades H158/H159 `portal-correlated`, target family unrecorded.
- **66**: llmwiki grades H66 `portal-correlated` — cloud-path evidence only, no local off-grid delta-test.

The previously-claimed mitigation — post-write readback — is structurally incapable of catching a wrong name→register mapping: a wrong-but-writable register is firmware-ACKed and reads back exactly the value written (#476).

- **Routing**: `local_write_blocked_reason` on `async_write_with_cloud_fallback` (`utils.py`), threaded through `_write_parameter` and `_write_voltage_register` — the local path is never attempted; the write lands via the existing cloud path (with the `note_parameters_written` cache seed on HYBRID) or, without a cloud client, raises a clear, actionable `HomeAssistantError`. Entities stay readable everywhere — local reads are verified and harmless.
- **Polarity (tribunal round 1, HIGH)**: the gate **fails CLOSED** via new `is_positively_non_offgrid_family()` — EG4_OFFGRID, a missing family, and UNKNOWN all degrade to the cloud-only route; local writes are permitted only for a positively resolved non-off-grid family (#476 precedent: an unidentified unit might BE an off-grid inverter). Entity-creation gates are unchanged (`is_offgrid_family` still fails open there — removing entities needs positive identification; refusing an unverified write needs the opposite).
- EG4_HYBRID keeps the reg 160 local-first route (hardware-verified: FlexBOSS21 fw FAAB-2727, read+write) and all local-first routes for 158/159/66 (regression-guarded).

### Sibling audit (final)
- **Regs 160/161 (SOC)**: cloud-only on off-grid/unresolved; 160 local-first on resolved EG4_HYBRID.
[...]

Issue #563, opened 2026-08-12
AC charge mode is a sensor not a switch


brendonlobo123:
### Integration Version

Beta10

### Home Assistant Version

2026

### Connection Mode

Hybrid (Local + Cloud)

### Device Model(s)

12000XP

### Describe the Bug

Ac charge mode is based on time and not on a switch. Setting times activates the mode. So it should be a sensor,? Binary? 

### Steps to Reproduce

Turn it off and it still follows the schedule. 
Maybe if it can clear the schedule or I could use automation or helper

### Affected Entity ID(s)

switch.garage_eg4_12000xp_inverter_ac_charge_mode

### Debug Logs (required)

[home-assistant_eg4_web_monitor_2026-08-12T09-36-38.826Z.log.txt](https://github.com/user-attachments/files/30975447/home-assistant_eg4_web_monitor_2026-08-12T09-36-38.826Z.log.txt)

### Diagnostics Download (required)

[config_entry-eg4_web_monitor-01KSWA215KHGP5HX9GJFN14SCF (2).json](https://github.com/user-attachments/files/30975453/config_entry-eg4_web_monitor-01KSWA215KHGP5HX9GJFN14SCF.2.json)

### Confirmation

- [x] I attached DEBUG-level logs captured while the problem was occurring (not just the default warnings log)
- [x] I attached the diagnostics download (or explained above why it is not possible)

### Screenshots

<img width="1080" height="2424" alt="Image" src="https://github.com/user-attachments/assets/0cdb83ef-571f-4a93-b4a1-f98d60e4ede0" />

brendonlobo123:
Might be similar for ac first mode too

Issue #558, opened 2026-08-09
reg 161 local write is live on EG4_OFFGRID but its stated mitigation (post-write readback) cannot detect a wrong mapping


btli:
Surfaced by the llmwiki tribunal (round 4) and verified against code before filing.

## What the code does

`ACChargeEndBatterySOCNumber` (`number.py:1436`) is created for `EG4_OFFGRID` (`number.py:660`) and writes via:

```python
await self._write_parameter(
    value,
    local_param=PARAM_HOLD_AC_CHARGE_END_BATTERY_SOC,   # local path offered
    cloud_write=lambda: _write_cloud_named_parameter(..., verify_register=161),
)
```

`async_write_with_cloud_fallback` (`utils.py:185`) documents: *"Local transport attached, link believed up: try `local_write` first"*. So on LOCAL and HYBRID installs the **local write is the default route**, not a fallback.

## The concern

The entity's own docstring states:

> LOCAL Modbus writes to reg 161 are hardware-UNVERIFIED on the off-grid family — all #331 write evidence is the cloud holdParam path. A silently-ignored local write is covered by the named write path's post-write parameter readback plus the HYBRID cloud fallback.

**A post-write readback cannot discharge this risk.** Per this repo's own #476 postmortem, a wrong-but-writable register is firmware-ACKed: no exception, no fallback, no log above DEBUG, and it reads back exactly the value written. If pylxpweb's name→register mapping for `HOLD_AC_CHARGE_END_BATTERY_SOC` resolves to the wrong address, the readback reads the *same wrong address* and reports success. The mitigation is structurally incapable of detecting the failure it is cited against.

This is the same reasoning that made reg 110 bit 8 ship wrong for an extended period (#476), and the same reasoning the AC Couple bit (#471/#472) is currently gated on.

## Not asserting a bug

No evidence that the mapping IS wrong — only that the stated safety argument does not hold, so the risk is currently **un-discharged rather than mitigated**. Options, in increasing cost:

1. Document it as an accepted un-discharged risk (cheapest; what the wiki will now say).
2. Route reg 161 cloud-only on `EG4_OFFGRID` until a local write is hardware-confirmed, matching how #471/#472 handles the unpinned AC Couple bit.
3. Discharge it properly: named vendor action on a known off-grid family, raw integer register read before and after, restoration — the full tuple.

Option 2 is what the codebase already does for the analogous unpinned case, which argues for consistency.

## Related
[...]

Issue #289, opened 2026-07-02
12000XP v2: control quirks — Off Grid Mode self-clears, Battery Backup Mode write rejected, AC Charge semantics unclear


btli:
Reported by @mjstrand in https://github.com/joyfulhouse/eg4_web_monitor/issues/259#issuecomment-4866 (2026-07-01), after confirming the #259 fix restored Controls on his 12000XP v2 (beta.17).

## Observed on 12000XP v2 (serial 61062J0147)
1. **Off Grid Mode switch clears itself** within ~10 seconds of being set.
2. **Battery Backup Mode** errors: "failed to enable working mode". Notably the EG4 maintenance page does not expose a working-mode control for the 12000XP v2 at all — the register may be rejected/unsupported on this family (compare #251's idle-write rejection pattern).
3. **AC Charge Mode** switch turns off and stays off, but the unit still powers the load from the grid — expected semantics need documenting (AC *charge* vs *passthrough*).
4. **Charge Last** stays on when set but produces no observable behavior change — may be inert on this family.

## Resources
- Reporter attached the EG4 maintenance Remote Set PDF for this unit: [Remote_Set_61062J0147-2.pdf](https://github.com/user-attachments/files/29573828/Remote_Set_61062J0147-2.pdf) — canonical list of what the portal itself exposes for XP v2.
- Reporter offered temporary portal + HA access for testing (see #259 thread).

## Likely direction
Family-aware control gating (the #259/#281 `is_supported_control_model()` pattern) may need per-family register capability maps for EG4_OFFGRID v2 — some HOLD-21/110 bits appear unsupported or auto-reverted by firmware on this family. Read-back-after-write verification would also surface firmware rejections honestly instead of leaving an optimistic switch lit.

## Notes
Post-3.4.0; his primary automation need is covered by the AC Charge schedule time entities (#277, shipping in beta.18).

mjstrand:
Just letting you know that these are really not issues for me. Sure it would be nice to have things disabled if it doesn't work with a specific model, but I think most everyone in the HA community knows to be cautious with settings. I did not do a thorough investigation of what happened at the inverter level, but if you want me to spend more time detailing which impact changes and which do not, I would be happy to do so. I owe you for getting the AC start and end time added in so quickly.

btli:
[...]

PR #600, opened 2026-08-29
audit: derive the #558 protected set from the #570 off-grid write-evidence sweep

### LINKED ISSUE #570: Audit write-evidence grades for every off-grid-writable register (derive the #558 protected set)
Follow-up from PR #569 tribunal (claude_code r2 LOW).

The protected-register set introduced by #558 (regs 158/159/160/161, then 66) grew by spot-checks, not derivation. Per the repo's own completeness rule ("a completeness claim is load-bearing — ask what derives the set"), sweep EVERY writable control the integration creates on EG4_OFFGRID against the llmwiki register-keeper write-evidence grades (local off-grid delta-test vs cloud/portal-correlated vs lineage-inferred), then either extend the cloud-only routing or record per-register evidence citations. Candidates explicitly flagged: PV charge power (reg 74), battery charge/discharge current numbers.

Origin: joyfulhouse/eg4_web_monitor#558, PR #569, tribunal round 2.

btli:
Fixes #570.

## The derivation (the point of the issue)

Frame: every writable control the integration creates on `EG4_OFFGRID`, derived from code, not memory — entity-creation blocks and family gates in `number.py` / `switch.py` / `select.py` / `time.py` / `button.py` / `services.py`, classified by write mechanism per the llmwiki README procedure (router / switch-action / direct library call / background). Result:

- **Number platform, local-first router** (`_write_parameter` / `_write_voltage_register`): regs **66, 158, 159, 160, 161** (already protected by #569) plus **74** (PV charge power), **101/102** (battery charge/discharge current), **105/125** (SOC cutoffs), **202** (stop-discharge voltage), **227/228** (system charge SOC/volt limits), **169/100** (cutoff voltages), **22** (PV start voltage — no ledger row at all). Cloud-only writers by construction: AC-Couple SOC pair, Smart Load numbers.
- **Outside the router** (recorded, not rerouted — see blind spots below): working-mode/function bit writes created on off-grid (H110 b3/b4/b14, H179 b11/b15, H21 b0 EPS, forced-charge enable), the H179 b9/b10 regime selects and H20 PV-input select, operating-mode via `inverter.set_operating_mode`, schedule time entities (H68–73, H152–157, H256–259 via packed FC06 `write_register`), and `QuickChargeDurationNumber`'s live H234 adjust (gated on a live local H233 b0 read the recorded off-grid boundary itself rejects). Suppressed on off-grid and therefore out of frame: H67, H82/H83/H84–89, H103, H116, H117, H206–H232 peak shaving, FUNC_AC_CHARGE (H21 b7, #563), FUNC_BATTERY_BACKUP_CTRL (H233 b1, #289), quick-charge switch pure-LOCAL (#569 fail-closed).

Criterion (unchanged since the #558 tribunal): protected when the llmwiki ledger records **no local off-grid delta-test**. Applied to the derived router set, **every register fails it** — so the protected set is now *all of them*, a derived rule instead of a spot-checked list. The #570 firmware verification proves H158–H161 are correctly mapped on CEAA/CCAA, which discharges the wrong-address risk there but does **not** upgrade them to local writes — the version-gated upgrade is recorded as a candidate on #570 only, per the issue scope.

## Routing change (minimal, existing mechanism)

`local_write_blocked_reason = _offgrid_cloud_only_reason(...)` extended to the eleven remaining router-driven scalars (three via `offgrid_local_write_unverified=True` on the voltage specs). Fail-closed polarity preserved: EG4_OFFGRID, missing family and UNKNOWN route cloud-only; pure-LOCAL raises the actionable error; positively resolved non-off-grid families keep local-first (regression-guarded). The derivation, and the mechanisms the gate cannot reach, are recorded in `_offgrid_cloud_only_reason`'s docstring.

## Evidence ledger (llmwiki)

- `40-hardware/registers.md`: H158/H159/H160 gain `firmware-proven` mapping rows (CEAA/CCAA, with proven raw constraints incl. **H160 min=1 not 0**) and `hardware-toggle-proven` tested-unit rows from the 2026-08-13 live sweep (raw valueFrame before/after/restore on FlexBOSS21 + 18kPV); H161 gains the firmware mapping row and the **second-unit inert-write confirmation** (anomaly resolved as the pre-documented grid-tied family quirk — C6 strengthened, still open); H66 gains the firmware writability/range row (semantics NOT verifiable — grade not upgraded) and the **pylxpweb 0..150 range conflict** (filed pylxpweb#272, fixed PR #273); H21 b7 `firmware-proven` inert (confirms #563); H233 boundary lineage-scoped (CEAA rejection `firmware-proven`, CCAA partial). Accounting: 336→345 counted claims, 32→41 proven; awk audit reproduces the subtotals.
- Falsification sweep: the pre-#569 "local-first, ungated" narratives in `data-semantics.md`, `controls-and-writes.md` (§1 router tree + §2.4), C7, the keeper's H233/H227 shipped-path notes and README's keeper cache updated to the shipped fail-closed routing; `log.md` entry appended.

## pylxpweb follow-ups (report only, no pylxpweb changes here)

Both addressed upstream: **#271** (H160 min 1) and **#272** (H66 raw ≤100) are CLOSED, fixed by **pylxpweb PR #273** (merged, with unit tests); release roll-up staged in draft PR #274.

## Gates

- `ruff check` + `ruff format`: clean
- `mypy --config-file tests/mypy.ini`: no issues in 47 files
- `pytest tests/`: **3128 passed, 9 skipped**
- Tier validators: silver / gold / platinum all exit 0

Tests: new `TestSweepExtendedProtectedRouting` matrix (cloud-never-local, pure-LOCAL raise per register, resolved-hybrid local-first guard, fail-closed unresolved family); ten pre-sweep local-first pins given resolved EG4_HYBRID scaffolds (the #569 protocol); the off-grid cutoff-voltage local-first pin deliberately inverted to the cloud route.


btli:
## Adversarial round 1 — Kimi MED adjudicated: CONFIRMED (and broader), fixed in `c89c98c`

**Adjudication against the code.** The finding is real, and the reachability is worse than stated: it does not even depend on a CCAA local H233 read. On `EG4_OFFGRID` + HYBRID, `is_quick_charge_active_live` routes through `_quick_charge_prefers_cloud` (the #296 mitigation, `coordinator_mixins.py`) to the **cloud** getStatusInfo — so a cloud-started quick charge reports active with **no local H233 read at all**, and `QuickChargeDurationNumber.async_set_native_value` then fired the **local** reg-234 write on either off-grid lineage. That was a designed path ("#296 round 2": `_read_offgrid_quick_charge_minute` deliberately mirrors reg 234 locally so the number's read and write sides agree), on a register with no off-grid write evidence (H234 `portal-correlated`). The docstring's rationale was therefore wrong twice: the gate it cited is cloud-routed on off-grid, and the H233 rejection is CEAA-scoped anyway — the classic lineage-scoped-negative-claim-generalized-to-the-family failure.

**Fix (`c89c98c`).** The live-adjust branch now requires `is_positively_non_offgrid_family` (same fail-closed polarity as the protected scalars); off-grid/unresolved families take the shipped CLOUD-branch behavior — the value is stored as the start preference and applied as the next cloud start's `minute` parameter, which is exactly what a CLOUD-mode off-grid install already did in the same situation. The local reg-234 **read** (off-grid HYBRID mirroring) is unchanged — reads carry no wrong-write hazard.

**Tests.** New `TestQuickChargeDurationOffgridLiveAdjust` in `tests/test_offgrid_write_routing.py`: off-grid / UNKNOWN / featureless + active charge → no local write, live check short-circuited, preference stored; resolved EG4_HYBRID regression guard keeps the live reg-234 write. Six pre-existing live-adjust pins given resolved `EG4_HYBRID` scaffolds. Shown RED against the unfixed code (the four family-less live-adjust pins failed before the scaffold fix — same signal, inverted).

**Wiki (the wiki follows the code).** Keeper H234 row now records the off-grid write-evidence absence and the shipped gate; `data-semantics.md` gate-limits note and `controls-and-writes.md` router row corrected; correcting entry appended to `llmwiki/log.md` (`## [2026-08-29] lint | QuickChargeDuration blind-spot rationale corrected...`) per the append-only rule.

**Gates re-run at `c89c98c`:** ruff check/format clean, mypy clean (47 files), pytest **3132 passed / 9 skipped**, silver/gold/platinum validators exit 0.

btli:
## Adversarial round 2 adjudicated — all 7 items fixed in `1f39ad1` + `8516c07`

1. **MED (Codex), H105 range — CONFIRMED.** Adjudication: the canonical H105 definition (pylxpweb `inverter_holding.py` address 105) and the cloud writer `set_battery_soc_limits` both enforce **10–90** (ValueError outside); the entity's 0–100 was the outlier — pre-PR the mismatch already existed on the pure-CLOUD path, and the sweep routing made it the only path on off-grid. Fix: `OnGridSOCCutoffNumber` now advertises, validates and reads 10–90 (`ONGRID_SOC_CUTOFF_MIN/MAX`), erroring clearly at the entity before any writer runs. Boundary tests added (10/90 land via the cloud writer; 9/91 raise `HomeAssistantError` with no writer call; H125's genuine 0–100 range regression-guarded). pylxpweb validation untouched.
2. **MED (Codex), H22 unledgered — CONFIRMED.** Added the keeper row: `portal-correlated` (canonical pylxpweb definition verified present at pin `ab87902` + the PR #359 verified named-volts cloud route), with two `asserted-unverified` sub-claims — pylxpweb's note that **reg 22 also carries LSP function bits** (an extra reason the local scalar write stays gated) and the 140 V firmware floor from the spec comment. Accounting 345→346 counted claims, 41 proven (11.8%), awk-reproducible. The "no ledger row" phrasing in README/code/tests updated.
3. **LOW (Codex), stale routing rationale — fixed.** `_offgrid_cloud_only_reason` now states the post-sweep keeper facts (H158–H160 `hardware-toggle-proven` on the tested hybrids via the cloud named path; H66 firmware-writable 0..100, semantics unverifiable) while keeping the unchanged criterion: no register in the set has a **local off-grid** delta-test.
4. **LOW (Codex+Grok), README worked example — fixed.** The blind-spot example now scopes the local H233 drive to positively resolved non-off-grid families and records the pre-#569 pure-LOCAL off-grid exposure as history closed by the fail-closed gate; the grep-invisibility lesson is preserved.
5. **LOW (Grok), §1.3 QCD contract — fixed.** The bypass table now records the family gate: off-grid/unresolved never call `is_quick_charge_active_live` and take the CLOUD-branch preference store; the None→raise contract is scoped to resolved non-off-grid families.
6. **LOW (Codex), pins — fixed** (`8516c07`). Per the #559 precedent, the four pages' front matter now names the PR #600 branch commit `1f39ad1` for claims describing this change set, to be re-pinned to the mainline merge SHA at the release cut; pre-existing claims keep their original pins.
7. **log.md** — round-2 correction entry appended (append-only preserved; the round-1 entries stand uncorrected in place with later correcting entries, per the log's own rules).

**Gates at `1f39ad1`:** ruff check/format clean, mypy clean (47 files), pytest **3138 passed / 9 skipped**, silver/gold/platinum validators exit 0 (`8516c07` is wiki-front-matter/log only).

btli:
## Adversarial round 3 adjudicated — both LOW fixed in `d4888cf`

1. **LOW, stale H158/H159 evidence comments (number.py:3069) — fixed, plus the whole-tree sweep.** The voltage-spec comments and — found by grepping every remaining evidence phrasing — the reg-66/160/161 entity docstrings and the routing-test module/class docstrings still carried pre-sweep wording ('only write evidence is a cloud delta-test, family unrecorded', 'no write evidence for H159 at all', 'FAAB-2727 exercise not yet ingested/graded, ledger still portal-correlated') that this PR's own keeper rows superseded. All now state the post-sweep facts: every H158–H160 write proof is CLOUD-path (`hardware-toggle-proven` on the tested FlexBOSS21/18kPV hybrids — which licenses no local write anywhere); H66 is firmware-writable 0..100 with unverifiable semantics and no write tuple; H160's FAAB-2727 charge-start behavior stays a CHANGELOG-recorded semantic observation, distinct from the graded write-path proof. Verified zero remaining hits for the stale phrasings across `custom_components/` and `tests/`.
2. **LOW, unformatted test file — fixed, root cause identified.** My gate runs executed `ruff format --check custom_components/` only, while CI runs `ruff format --check .` repo-wide with pinned ruff **0.15.5** — so `tests/test_offgrid_write_routing.py` (and this round's docstring edits) slipped through. Re-ran the gate exactly as CI does (`uvx ruff@0.15.5`, `--isolated`, repo root): `ruff format --check .` → **140 files formatted**, `ruff check .` → clean. Repo-wide 0.15.5 is now part of my pre-push loop for this branch.

**Gates at `d4888cf`:** ruff 0.15.5 check + format clean repo-wide, mypy clean (47 files), pytest **3138 passed / 9 skipped**, silver/gold/platinum validators exit 0. No llmwiki changes needed this round (the corrections were code/test docstrings catching up to the keeper — the wiki was already right).

btli:
## Adversarial round 4 adjudicated — 3 MED + 2 LOW, all fixed in `ce72b73`

1. **MED, quick-charge switch on unresolved families — CONFIRMED.** Adjudication: `_prefers_cloud_control` was `is_offgrid_family(...) and has_http_api()` — positive off-grid only — so an UNKNOWN/missing family with cloud creds + local transport still invoked pylxpweb's local-first paired H233/H234 write before fallback. #569's recorded rationale for that ("other/unresolved families keep local-first-with-cloud-fallback — safe, the fallback works") predates the CCAA firmware verdict: on CCAA the H233 write is **silently ACCEPTED** with unproven bit-0 semantics, and a silent ACK never triggers a fallback (the #476 mechanism). The switch has no other family gate covering this configuration (`_offgrid_without_cloud` only guards the cloud-less case), and the QCD preference-store implied a cloud start the gate did not guarantee. Fix: the predicate now **fails closed** — `not is_positively_non_offgrid_family(...) and has_http_api()`; resolved non-off-grid families keep local-first (regression-guarded). Tests: off-grid/UNKNOWN/featureless + cloud + local transport → cloud start/stop endpoints, `enable_quick_charge`/`disable_quick_charge` never called.
2. **MED, H66 ceiling — CONFIRMED, family-scoped fix.** Adjudication: pylxpweb #273 capped only the **canonical definition** (10 kW); the cloud method `set_ac_charge_power` still accepts 0–15 and the local transport RMW validates nothing — so the entity's bound is the effective gate. The firmware proof of the raw-100 cap is CEAA/CCAA-scoped; hybrid acceptance of 15 kW is shipped status quo (DATA_MAPPING raw/UI 0-150 = 0-15 kW) with no firmware proof either way — per the family-scoping instruction I did NOT narrow globally. `ACChargePowerNumber.native_max_value` is now dynamic: **10 kW on off-grid/unresolved (fail closed), 15 kW on positively resolved non-off-grid**; the read window tracks the bound (over-ceiling values read unknown instead of tripping HA's out-of-range state error). Boundary tests: off-grid 10.0 lands via cloud, 10.1 raises before any writer; hybrid keeps 15.0 → raw 150 local.
3. **MED, H160 floor — CONFIRMED, same family-scoped treatment.** `native_min_value` is dynamic: **1 on off-grid/unresolved** (0 → exception 03, firmware-proven; pylxpweb #273 set canonical min 1), **0 on positively resolved non-off-grid** with the documented rationale (0's semantics there unverified but nothing proves it invalid — revisit if a hybrid image is decoded). Read window tracks the floor (a factory-programmed 0, which bypasses the Modbus writer, reads unknown rather than erroring). Boundary tests: off-grid 1 lands via holdParam, 0 raises; hybrid 0 still writes local-first; 0 added to the out-of-range parametrization.
4. **LOW, QCD class docstring** — now states the fail-closed family gate, the cloud-routed active check, and the preference-store behavior explicitly.
5. **LOW, `_read_offgrid_quick_charge_minute` docstring** — corrected to **mirror-read-only** with an explicit warning: do not "restore" read/write symmetry by pointing the setter back at reg 234 — that reopens exactly the ungated off-grid local write this audit closed.

**Wiki (follows the code):** keeper H233 boundary routing row, `data-semantics.md` and `controls-and-writes.md` §2.4 updated for the extended fail-closed routing; round-4 `llmwiki/log.md` entry appended.

**Gates at `ce72b73`:** ruff 0.15.5 check + format clean repo-wide (`.`, `--isolated`), mypy clean (47 files), pytest **3147 passed / 9 skipped** (+9 new), silver/gold/platinum validators exit 0.

btli:
## Adversarial round 5 adjudicated — 2 MED + 3 LOW fixed in `7400342` + `050511d`, plus the demanded derivation pass

1. **MED (Grok), coordinator read-side gate — CONFIRMED.** `_quick_charge_prefers_cloud` still failed open (`is_offgrid_family`) while round 4's switch predicate failed closed, so UNKNOWN + cloud + transport cloud-STARTED a charge and then status-polled it via the local H233/H234 detail read — the #296 invisibility bug reintroduced on exactly the population round 4 protected. Fixed fail-closed (`not is_positively_non_offgrid_family(...) and client and transport`); read-side test pins UNKNOWN + cloud + transport → cloud status, `get_quick_charge_detail` never awaited. The reg-234 mirror read stays (reads are harmless; its docstring already carries the do-not-restore-symmetry warning).
2. **MED (Codex), fail-open creation window — CONFIRMED, fixed at the setters.** Adjudication: creation gates on the family predicate, not model strings, and deliberately fails OPEN — `test_xp_model_without_family_fails_open` pins that a 12000XP model with no family keeps H67/H82/H83/etc. by design (#259/#219: suppression needs positive ID). Closing CREATION would strip working controls from every not-yet-resolved hybrid, so per the 'or at minimum' branch the SETTERS now fail closed: **H67, H82, H83, H103, H116** route cloud-only on off-grid/unresolved, and the RAW **H117** write (no cloud path exists) is refused outright. First-run-UNKNOWN tests (cloud write lands, zero local writes; H117 raises) + resolved-hybrid regression guards (incl. the raw H117 write). H206 needed no gate — it was already cloud-only by construction.
3. **Derivation pass (so round 6 doesn't find a fourth):** classified every `is_offgrid_family(` site. Creation/suppression gates, Repairs flags, off-grid read blocks and read-only sensors/buttons stay fail-open by design (documented per site). One additional write path surfaced that no reviewer flagged: **schedule time entities** wrote packed FC06 registers local-first on any family, while the clear-schedule button already declared local off-grid schedule writes unsanctioned (#563) — `time.py` now passes the same fail-closed `local_write_blocked_reason` (off-grid/unresolved → classic cloud field writes; pure-LOCAL raises; resolved non-off-grid keeps local-first), with routing tests. Remaining ungated write surfaces are the bit-level switch/select writes and direct library calls — recorded in `_offgrid_cloud_only_reason`'s docstring and C7, deliberately not converted.
4. **LOW pins (Codex+Grok) — fixed in `050511d`.** Change-set claims re-pinned to `7400342` (contains every round's predicates); the front-matter comments now state the pin FOLLOWS the branch head during review and gets re-pinned to the mainline merge SHA at the release cut (#559 precedent) — the defect was citing a commit that predated a later round's formula.
5. **LOW (Codex), setup comment — fixed**: the QCD creation comment now states the family-gated live reg-234 write and the preference-store behavior for off-grid/unresolved.

**Wiki:** keeper schedule-write-boundary row, README H117 cells (now three exposure-narrowing facts), controls-and-writes router row, data-semantics gate table/limits, round-5 `log.md` entry (records the derivation pass and why creation stays fail-open).

**Gates at `7400342`:** ruff 0.15.5 check + format clean repo-wide, mypy clean (47 files), pytest **3160 passed / 9 skipped** (+13 new), silver/gold/platinum validators exit 0 (`050511d` is wiki front matter only).

btli:
## Adversarial round 6 adjudicated — 5 MED + 3 LOW, all fixed in `dd6f61e`, plus the demanded wheel sweep

1. **MED (Grok), H206 — CONFIRMED, and the derivation's method was the defect.** The r5 "cloud-only by construction" claim came from the ENTITY docstring ("cloud-write-only ... local name-writes are never used"); the pinned wheel falsifies it: `set_grid_peak_shaving_power` (base.py) is **transport-first onto raw H206** (deci-kW) whenever a transport is attached and the link is up — and the wiki's own README inventory had recorded exactly that while the derivation trusted the code comment. Fixed: the entity reaches the library method only on a positively resolved non-off-grid family; off-grid/unresolved write the cloud named parameter (`_12K_HOLD_GRID_PEAK_SHAVING_POWER`) directly — the wheel's own cloud leg. First-run tests (UNKNOWN/featureless 12000XP → cloud-named, library method never awaited) + resolved-hybrid guard. Entity/derivation docstrings corrected; wiki H206 rows updated.
2. **Demanded wheel sweep — complete.** Every cloud path the number/time gates rely on was verified against the pinned wheel: `set_battery_soc_limits`, both battery-current setters (direct `client.api.control` calls), every `_write_named_parameter` consumer (66/74/67/82/83/202 — that helper is pure `client.api.control.write_parameter`), `set_feed_in_grid_power_kw` and `set_system_charge_soc_limit`. **H206's method was the only transport-first one.** The verdict list is recorded in `_offgrid_cloud_only_reason`'s docstring and the round-6 log entry.
3. **MED (Grok+Codex), schedule revert — CONFIRMED, one fix for both.** The cloud-routed schedule write now passes `local_values` (the packed value under every alias key the decoder reads) so `note_parameters_written` seeds the local-raw cache. Convergence test: off-grid + stale local register holding 08:00, cloud-ack of 06:30 → no revert, entity shows 06:30 with the optimistic value cleared.
4. **MED (Codex), H158/H159 ranges — CONFIRMED.** New spec-level `offgrid_min/max_value` + dynamic `native_min/max_value` on `EG4VoltageNumber`: off-grid/unresolved advertise/validate/read **38.4–57.0 V (H158)** and **48.0–59.0 V (H159)**; resolved non-off-grid keeps the shipped 38–60 V. Boundary tests (57/48 land via cloud raw write; 58/47 and 38/60 fail clearly at the entity; hybrid 60 V regression-guarded). The cross-register H158≤H159 constraint stays firmware-enforced only, with the documented rationale: validating against a stale cached sibling produces false rejections.
5. **MED (Codex), H161 floor — CONFIRMED.** Dynamic floor 20 on off-grid/unresolved (0 kept on resolved non-off-grid, where the keeper records the register inert anyway); read window tracks it. Boundary tests (20 lands via holdParam, 19 raises); ≥H160 stays firmware-enforced per the same rationale.
6. **LOW, H117 message** — no longer promises a cloud route that does not exist; it names family resolution as the remedy, with the test pinning the new wording.
7. **LOW, stale FC06 claims** — `time.py` module docstring and controls-and-writes §6 corrected for the family-routed schedule writes.
8. **LOW, pins** — all change-set pin comments now use **branch-head language with no embedded SHA** (two successive pinned SHAs staled inside the review train — the root cause, not a one-off), keeping the explicit release-cut re-pin step.

**Gates at `dd6f61e`:** ruff 0.15.5 check + format clean repo-wide, mypy clean (47 files), pytest **3169 passed / 9 skipped** (+9 new), silver/gold/platinum validators exit 0.

btli:
## Adversarial round 7 adjudicated — the convergence class fixed as one mechanism, in `9f77c22`

**The class (3 MEDs).** Every write this PR re-routed through the cloud races the post-write refresh against portal→dongle→register propagation, and three gaps let the stale register win. Fixed systematically, not per-site:

1. **Coordinator settle window** (new `PARAMETER_WRITE_SEED_SETTLE`, 30 s): `_reconcile_parameter_read` previously "confirmed" a seeded key at a **disagreeing** fresh observation — the successful post-write refresh retired the seed at the pre-write value and optimism cleared onto it (Codex's schedule case verbatim). Now a disagreeing observation inside the window keeps the seed; agreement confirms immediately (existing grace); past the window fresh reads are authoritative again. **RED-proven on the real coordinator** (`TestParameterSeedSettleWindow`, stash-check: 2 failed pre-fix / green post-fix). The two existing tests that pinned immediate post-write authority now expire the window for their convergence phases — their eventual-convergence intent is preserved.
2. **Read-side gate** (Grok's snap-back): `_read_param_value`'s shipped inverter-first order let the pylxpweb attribute (refreshed from the same stale register) shadow the seed entirely. New `has_active_parameter_write_seed(serial, key)` on the coordinator; while a seed is live the read behaves params-first for that key (strict-`is True` so auto-mocked scaffolds keep their declared order — pinned both polarities in `TestActiveWriteSeedReadGate`).
3. **H206 branch seeding**: the r6 cloud-named branch never called `note_parameters_written` — it now seeds the acknowledged kW value (asserted in the first-run test).

**Mechanism sweep (one derivation):** router `local_values` seeding covers 66/74/101/102/105/125/158/159/160/161/169/100/22/202/227/228 and r5's 67/82/83/103/116; H206 seeds at its entity; schedule times seed every decoder alias key; H117 (refused, no cloud path) and QCD (preference store) write nothing cloud-side to converge. Per Grok's point that routing tests pass without convergence, the routing tests now assert **`native_value` after every cloud-routed write** against an emission-faithful mock (its contract pinned by the real-coordinator settle tests), and the time convergence test's refresh now actually writes the stale value into the cache before the seed overlay (Codex item 2) — RED without `local_values`.

4. **MED, H158 boundary** — the 38.4 V advertised floor on a whole-volt entity let HA accept a boundary the validation then rejected. Advertised off-grid bounds are now the whole-volt values inside the firmware windows (H158 **39–57**, H159 48–59), every advertised boundary writes as-is, and **both ends are boundary-tested for H66/H158/H159/H160/H161** (min and max land with exact writer payloads; min−1/max+1 fail clearly).
5. **LOW, symmetry comment** — `_read_quick_charge_status`'s "read and write sides agree" corrected to mirror-read-only with a pointer at the do-not-restore-symmetry warning; grepped the tree for remaining aligned/symmetry claims about H234/schedules — none left.
6. **LOW, durable pins** — change-set claims now cite **PR #600** (the diff survives branch deletion), and the release-cut re-pin is recorded as a **REQUIRED POST-MERGE ACTION** in `llmwiki/log.md` (round-7 entry is the tracking record).

**Gates at `9f77c22`:** ruff 0.15.5 check + format clean repo-wide, mypy clean (47 files), pytest **3174 passed / 9 skipped**, silver/gold/platinum validators exit 0.

btli:
## Adversarial round 8 adjudicated — 1 MED + 1 LOW, fixed in `d313f0d` (Grok CLEAN)

1. **MED, serial-wide settle stamp — CONFIRMED, fixed per-key.** The r7 window keyed on the inverter's newest write, so writing H102 re-armed H101's stale-read protection (masking a fresh external H101 change) and repeated sibling writes could extend the mask indefinitely. Now each seeded key carries its **own write stamp** (`_parameter_write_seed_key_stamps`, cleaned on every retirement path — reconcile, confirmed-grace expiry, per-key overlay expiry, whole-serial TTL), **retires on its own confirmation** (an agreeing observation ends the protection, so a confirmed key's later external changes stay visible even while siblings are being written), and **expires on its own 30 s window** regardless of sibling writes. Tests pin both demanded cases — (a) H102 written after H101 confirmed → external H101 change visible immediately while H102's own fresh seed holds; (b) sibling writes landing as H101's window expires do not extend its mask — both **RED against the r7 serial-wide code** (stash-check: 2 failed) and green now; every r7 convergence pin (settle-keeps-seed, agreement-confirms, post-window authority, entity read-gate, routing native_value assertions, time convergence) stays green.
2. **LOW, front matter — CONFIRMED, the r7 fix had stopped one level short.** The durable-pin language lived only in the front-matter *comments* while the machine-read `verified-against:` values still named pre-PR commits. All four edited pages (`registers.md`, `controls-and-writes.md`, `data-semantics.md`, `README.md`) now carry the compound pin value — `"<pre-PR SHA> + PR #600 (change-set claims; re-pin to the merge SHA at the release cut)"` — so the pin itself licenses the PR-#600 claims; the release-cut re-pin remains a tracked REQUIRED POST-MERGE ACTION in `llmwiki/log.md` (r7 entry), with a round-8 entry recording this correction.

**Gates at `d313f0d`:** ruff 0.15.5 check + format clean repo-wide, mypy clean (47 files), pytest **3176 passed / 9 skipped** (+2 new), silver/gold/platinum validators exit 0.

btli:
## Adversarial round 9 adjudicated — 1 MED + 2 LOW, fixed in `6c42b61` (Grok CLEAN)

1. **MED, no follow-up read at settle expiry — CONFIRMED, fixed with the existing machinery.** A retained disagreement now schedules a **one-shot targeted per-device refresh** at that key's settle expiry (`_schedule_seed_settle_recheck`): deduplicated per serial, reusing `async_refresh_device_parameters` (no new polling loop; at most one extra read per disagreement episode, respecting the read-budget conventions), cancelled on coordinator shutdown. The deferred refresh reconciles through the normal pipeline — past the window its authoritative observation retires the seed to device truth, so a never-propagating cloud write or an immediate external change resolves within ~one settle window instead of the hourly poll; keys still inside their own windows re-arm the recheck naturally from that refresh's reconciliation. **Test (RED against pre-r9 via stash-check):** disagreeing read inside the window → recheck scheduled → `async_fire_time_changed` fires the timer → the targeted refresh runs and its still-stale-but-now-authoritative observation resolves the seed (cache and seed at the observed value), with **no manual reconciliation calls**. All r7/r8 settle pins stay green.
2. **LOW, canonical seeding docs** — `data-semantics.md` §7 (the canonical parameter-seeding section) gains four rows: the disagreement settle window, per-key stamps + per-key confirmation retirement, the settle-expiry recheck, and the params-first override while a seed is active — each cited to its coordinator/number symbol as part of the PR #600 change set.
3. **LOW, front-matter schema — CONFIRMED; the r8 fix had broken the schema it served.** Read `_conventions.md` first as instructed: `verified-against:` admits exactly two forms, bare commit or multi-repo labelled commits — no sanctioned mechanism for in-flight change-set claims. All four pages' pins are back to **bare pre-PR commits** (`e9853eb`/`e146d91`/`9f6d6e2`×2); the PR-#600 claims are licensed **per claim by their inline PR #600 / issue #570 citations** (the r7 comment-level form the conventions permit), the front-matter comments state that form explicitly, and the release-cut re-pin remains the tracked REQUIRED POST-MERGE ACTION in `llmwiki/log.md` (r7 entry; r9 entry records this correction — 'a citation fix that itself broke the citation schema').

**Gates at `6c42b61`:** ruff 0.15.5 check + format clean repo-wide, mypy clean (47 files), pytest **3177 passed / 9 skipped** (+1 new), silver/gold/platinum validators exit 0.

btli:
## Adversarial round 10 adjudicated — 1 MED + 3 LOW, fixed in `c7c01e7` (Codex gpt-5.6-sol; grok crashed, kimi pending)

1. **MED, recheck dedup loses the earliest per-key deadline — CONFIRMED, fixed.** The r9 dedup keyed on the serial alone and kept whichever recheck scheduled *first*: if a later-expiring key's disagreement scheduled before an earlier-expiring key's, the earlier key's retained seed masked a genuine external change until the later window ran out. `_schedule_seed_settle_recheck` now tracks the pending deadline per serial (`_parameter_seed_recheck_at`) and keeps the **minimum**: a call with an earlier per-key expiry cancels the pending timer and pulls the one-shot forward; a later one still dedups away (its keys resolve or re-arm from the earlier shot's own reconciliation, as r9 documented). **Test (RED against pre-r10):** two keys with staggered write stamps, the later-expiring key's disagreement reconciled first (recheck pending at ~+26s), then the earlier key's (~+6s) — the deadline is pulled forward and `async_fire_time_changed` at +10s fires exactly one targeted refresh; under the old dedup nothing fires until +26s. Test-construction note: patching `time.monotonic` warps asyncio's loop clock (and therefore the very `async_call_later` delay under test), so the test backdates the write stamps instead of warping "now".
2. **LOW, HA stop path leaves recheck timers live — CONFIRMED, fixed.** `_async_handle_shutdown` (the HA-stop-without-unload path) never cancelled `_parameter_seed_recheck_unsub` — only `async_shutdown` did — so a pending recheck could fire its targeted refresh after the transports and cloud session were shut down. Both teardown paths now share `_cancel_seed_settle_rechecks()`, which also clears the new deadline map.
3. **LOW, routing comment + mirrored wiki claim overstate "classic cloud field writes" — CONFIRMED, corrected.** The `time.py` local-block comment and the mirrored schedule-write routing row in `llmwiki/40-hardware/registers.md` both said off-grid/unresolved schedule writes route through the classic cloud field writes — false for the writeTime families (Generator/Off-Grid/Peak Shaving, which is exactly what an off-grid inverter's Generator Charge schedule uses): those take the atomic `write_time_parameter` portal call; only the classic families (AC Charge/First, Forced Charge/Discharge) take the per-field hour/minute writes. Both statements now describe the per-schedule routing (`_async_write_cloud` is the dispatcher either way — the fail-closed *local* gate is unchanged). Round-10 `llmwiki/log.md` entry appended in the exact heading form.
4. **LOW, H206 contract comment contradicts the shipped split — CONFIRMED, corrected.** `const/modbus.py`'s `PARAM_HOLD_GRID_PEAK_SHAVING_POWER` comment still said the raw encoding was "presumed deci-kW but unverified … cloud-write-only", contradicting the r6 routing this PR ships and `number.py` documents. It now states the family-scoped split: resolved non-off-grid families go through pylxpweb's transport-first raw-H206 method (deci-kW hybrid-verified, pylxpweb#158 raw 41 → 4.1 kW, cloud named fallback); off-grid/unresolved families are cloud-named-parameter-only (no off-grid H206 write evidence). The "never via the local transport name map" clause survives — that part was and is true (the old name-map entry pointed at unrelated reg 231).

**Gates at `c7c01e7`:** ruff 0.15.5 check + format clean repo-wide, mypy clean (47 files), pytest **3178 passed / 9 skipped** (+1 new), silver/gold/platinum validators exit 0.

btli:
## Adversarial round 11 adjudicated — 2 LOW, fixed in `2e60430` (Codex gpt-5.6-sol; grok unavailable this round)

1. **LOW, recheck can re-arm during teardown — CONFIRMED, fixed.** Both teardown paths cancel the settle-recheck timers *before* the awaited base-class teardown, and `_schedule_seed_settle_recheck` had no shutdown guard — so an in-flight reconciliation completing during that await could re-arm a timer whose deferred refresh would then run against closed transports/cloud session. `_cancel_seed_settle_rechecks()` now also **latches scheduling closed** (`_parameter_seed_recheck_closed`, one-way — both callers, `async_shutdown` and `_async_handle_shutdown`, are terminal for the coordinator instance), and `_schedule_seed_settle_recheck` is a no-op once the latch is set. **Test (RED-verified against pre-r11):** a disagreeing reconciliation injected mid-`async_shutdown` (during the awaited cloud-session close, after the cancel ran) → no timer in either recheck map afterward, and firing time +120s produces **zero** refresh calls; without the latch the re-armed timer survives and the assertion fails.
2. **LOW, H206 seeding comment's false name-map rationale — CONFIRMED, corrected (wheel-verified).** The cloud-branch seeding comment in `GridPeakShavingPowerNumber.async_set_native_value` justified the kW seed with "the local name map never carries this key". Checked against the pinned wheel: `pylxpweb/constants/registers.py` maps `206: ["_12K_HOLD_GRID_PEAK_SHAVING_POWER"]` and the key is in `LOCAL_PARAM_SCALE_DIV10`, so the transport decode divides raw deci-kW by 10 and a **local read surfaces kW** — the same engineering value the cloud returns. The seeded **kW value was already correct** — kW is the unit the parameter cache carries for this key on every path, which is exactly what `note_parameters_written` overlays — but the false rationale could have motivated a future "fix" seeding raw deci-kW, a 10× state defect. The comment now states the actual mapping/decode and why kW is the invariant unit. Swept for reoccurrence: no other file repeats the claim, and the wiki's H206 rows (`40-hardware/registers.md`) already say 0.1 kW raw — no wiki content change; round-11 `llmwiki/log.md` entry appended.

**Gates at `2e60430`:** ruff 0.15.5 check + format clean repo-wide, mypy clean (47 files), pytest **3179 passed / 9 skipped** (+1 new), silver/gold/platinum validators exit 0.

btli:
## Adversarial round 12 adjudicated — 1 MED, fixed in `e137d33` (Codex gpt-5.6-sol; kimi quota-exhausted again)

1. **MED, fired recheck runs unowned parameter I/O across teardown — CONFIRMED, fixed with the existing ownership machinery.** The fired settle-recheck callback popped its only coordinator-owned cancellation handle and then awaited `async_refresh_device_parameters` with no owner — so a timer firing just before reload/shutdown left the parameter I/O running against detached transports or a closing cloud session, racing the replacement coordinator or delaying HA shutdown. Two changes inside `_schedule_seed_settle_recheck`, no new mechanism invented:
   - **Latch re-checked AFTER the fire:** the timer callback is now a sync `@callback` that re-checks the r11 closed latch (plus the mixin's `_background_scheduling_stopped` producer gate) before starting any I/O — covering the fire-vs-shutdown ordering where the callback outraces teardown's cancel.
   - **The refresh is coordinator-owned:** it runs through the repo's existing `BackgroundTaskMixin` convention — `hass.async_create_task` + `_background_tasks` + the `_remove_task_from_set` / `_log_task_exception` done callbacks — so `_cancel_background_tasks` (which **both** teardown paths already execute *after* `_cancel_seed_settle_rechecks` sets the latch, in `_async_shutdown_work` and `_async_handle_shutdown_work`) cancels and awaits it. The ordering latch-first-then-drain means no new task can slip in behind the drain.

   **Tests (both RED-verified against pre-r12 via patch-revert):** (a) timer fires, the refresh blocks in flight → `async_shutdown` cancels and awaits it (`CancelledError` observed inside the refresh double, `_background_tasks` drained, shutdown completes); (b) recheck armed, latch set as if teardown began, timer fires → zero refresh calls and no task created.

**Gates at `e137d33`:** ruff 0.15.5 check + format clean repo-wide, mypy clean (47 files), pytest **3181 passed / 9 skipped** (+2 new), silver/gold/platinum validators exit 0.

btli:
## Adversarial round 13 adjudicated — 5 LOW (no behavior defects), fixed in `20eb767` (Codex gpt-5.6-sol)

1. **LOW, quick-charge status routing docstring — CONFIRMED, corrected.** `_read_quick_charge_status` still described the pre-r5 gate: "except for EG4_OFFGRID + HYBRID where register 233 is firmware-rejected". Both halves were wrong against the code below it: the gate is `_quick_charge_prefers_cloud`'s **fail-closed, unresolved-inclusive** predicate (any HYBRID not positively resolved non-off-grid), and the H233 rejection is **CEAA-scoped** (ILLEGAL DATA ADDRESS, #296) while CCAA implements the register with unproven semantics. A maintainer following the old wording could have restored the positive-off-grid-only gate that r5 removed. The docstring now states both correctly and defers to the predicate.
2. **LOW, H160 writer comment calls grid-tied cloud writes untested — CONFIRMED, corrected.** Stale since the #570 live sweep (2026-08-13) hardware-toggle-proved the named cloud path on **two** grid-tied units (FlexBOSS21 and 18kPV, raw 5→6→5; scope limited to the tested units — matching the `hardware-toggle-proven` row in `40-hardware/registers.md`). The `verify_register=160` readback **stays**, now with the correct justification: sibling H161 returned success while raw stayed 0 on the same units, so acknowledged-but-unapplied remains a live failure mode.
3. **LOW, inert attempt-then-fallback regression test — CONFIRMED, re-armed.** `test_fallback_write_seeds_cache_with_local_raw_value`'s featureless scaffold resolves as UNKNOWN family, which since r5 takes the blocked-local cloud route — the mocked local failure was never invoked and the test passed while pinning nothing. Re-scaffolded on `EG4_HYBRID` with an explicit `write_named_parameter.assert_awaited_once()` pinning that the local attempt ran and failed first. **Mutation-verified:** making the fallback always re-raise in `utils.py::async_write_with_cloud_fallback` turns exactly this test RED.
4. **LOW, inert link-down short-circuit test — CONFIRMED, re-armed.** Same mechanism for `test_hybrid_known_link_down_prefers_cloud_immediately` (time entities): re-scaffolded on `EG4_HYBRID`. **Mutation-verified:** neutering the `is_transport_link_down` short-circuit turns exactly this test RED (the local write is attempted and succeeds, so the asserted 2 cloud field writes never happen).
5. **LOW, off-grid routing tests cover only the classic leg — CONFIRMED, added.** New `test_offgrid_gen_charge_routes_atomic_cloud_write` (off-grid + unresolved parametrized): a Generator Charge write on the fail-closed route issues **no local FC06 to H256**, takes the **atomic `write_time_parameter`** call (`HOLD_GEN_START_TIME_1`, hour+minute together), and never touches the classic per-field `write_parameter` path. The routing class docstring also drops the same "classic cloud field writes" overstatement r10 corrected in `time.py` and the wiki.

Round-13 `llmwiki/log.md` entry appended (records the round-3/4 lesson: a routing change that adds an earlier branch can silently strand downstream tests on the new branch — green while pinning nothing).

**Gates at `20eb767`:** ruff 0.15.5 check + format clean repo-wide, mypy clean (47 files), pytest **3183 passed / 9 skipped** (+2 new), silver/gold/platinum validators exit 0.

btli:
**Final attestation quorum (maintainer-approved reduction):** Codex gpt-5.6-sol (xhigh) attested **CLEAN** on the final head `20eb767` after 14 adversarial rounds. Grok 4.6 last attested CLEAN at round 9 (`6c42b61`) before its usage balance was exhausted; Kimi K3 last attested CLEAN at round 2 (`8516c07`) before hitting its weekly quota. The maintainer approved merging on the Codex final-head attestation with this disclosure. Post-merge required action (llmwiki re-pin to the merge SHA) is tracked in `llmwiki/log.md`.

